PR #6, opened 2019-06-28
Support prettier's singleQuote option

### LINKED ISSUE #3: Configurable quotes
Use prettier's [`singleQuote`](https://prettier.io/docs/en/options.html#quotes) to correctly quote the attributes.
This should be overwritable via:
```json
{
  "overrides": [
    {
      "files": "*.pug",
      "options": {
        "parser": "pug",
        "singleQuote": true
      }
    }
  ]
}
```

Shinigami92:
closes #3 

@rdhar: I just found out that prettier's override option applies to the options by default.
At first I wanted to use the invers of `singleQuote` so JS code will be quoted the right way in pug files, BUT this means that you need to configure `singleQuote: false` in prettier's override for pug to use actually single quotes 🤔  

What do you think should be the default behaviour?!

On [pugjs.org](https://pugjs.org) there are used single quotes for attributes.
I for myself use double quotes by default because I'm used to it from html.

🤔

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,3 +11,5 @@ insert_final_newline = true
 [*.md]
 max_line_length = off
 trim_trailing_whitespace = false
+indent_style = space
+indent_size = 2
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -6,5 +6,14 @@
 	"singleQuote": true,
 	"tabWidth": 4,
 	"trailingComma": "none",
-	"useTabs": true
+	"useTabs": true,
+	"overrides": [
+		{
+			"files": "*.md",
+			"options": {
+				"tabWidth": 2,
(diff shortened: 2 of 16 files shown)


rdhar:
Hi @Shinigami92, sorry for the delayed response.

Since Pug.js is a template engine powered by JavaScript, it makes sense for it to adopt JavaScript's default Prettier configuration for details like this. As such, it should default to double-quotes.

Personally—like you—I always use double-quotes out of habit but rely on `"prettier.singleQuote": true` to always convert them to single-quotes for a "cleaner" look.

I believe that's what you've done in the latest commit for this PR so that's a double 👍👍 from me, thanks!

Shinigami92:
@rdhar Perfect, so we are on the same result

PR #37, opened 2019-11-10
chore: prettier should only be peer and dev dependence

JounQin:


JounQin:
@Shinigami92 Done. You don't need to translate the texts. 🤣

Shinigami92:
> Besides, I'd like to recommend you to use [commitlint](https://github.com/conventional-changelog/commitlint) which recommends prefix a meaningful type for commit messages.

I know commitlint a bit through other projects, but I do not like it that much for personal reasons because it blocks commit and push via VSCode GUI

PR #100, opened 2020-07-07
Support bracketSpacing option (only text-token)

### LINKED ISSUE #98: Option to not format double curly braces (mustache)?
## Request / Idea

I'm using [Stein](https://steinhq.com/) for a very simple website together with pug. Stein [Expedite](https://docs.steinhq.com/expedite-display-data) allows users to use column-names directly in HTML, as in `{{colname}}`. However, prettier seems to format this into `{{ colname }}` even when `bracketSpacing` is set to false. And Stein does not work if formatted that way. Is this possible, or am I doing something wrong / missing something?
My `.prettierrc.js` looks like below:

```js
module.exports = {
  singleQuote: true,
  semi: false,
  overrides: [
    {
      files: '*.pug',
      options: {
        parser: 'pug',
        bracketSpacing: false, // for stein expedite
      },
    },
  ],
}
```

## Input

```pug
div(
  data-stein-url='https://api.steinhq.com/v1/storages/[spreadsheet]/[sheet-name]',
  data-stein-limit='100'
) 
  div {{date}}

```

## Expected Output

exactly the same as input.

Shinigami92:
closes #98

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,23 @@ yarn prettier --write "**/*.pug"
 
 ### Prettier Options
 
+- `bracketSpacing`  
+  If you want to configure different bracketSpacing for pug than for js code, you can use prettier's override.  
+  _(Only text tokens are currently supported. If you want to expand the bracketSpacing option, please open a new issue.)_
+  ```json
+  {
+    "bracketSpacing": true,
+    "overrides": [
+      {
+        "files": "*.pug",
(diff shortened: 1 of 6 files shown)


Shinigami92:
@xerroxcopy Can you give me more examples for testing purposes?

Also embedded js and so on? Or just text tokens?

PR #106, opened 2020-09-14
Add pug aliased options

### LINKED ISSUE #105: New rule names of aliases for prettier defaults
## Request / Idea
Use new names or aliases for rules that are specific for Pug instead of using default ones.

## Current Configuration

```json
{
  "bracketSpacing": true,
  "overrides": [
    {
      "files": "*.pug",
      "options": {
        "parser": "pug",
        "bracketSpacing": false
      }
    }
  ]
}
```

## Expected Configuration

```json
{
  "bracketSpacing": true,
  "pugBracketSpacing": false
}
```

## Additional Context

Using files override won't work in mixed files, like Vue's Single File Components since it doesn't recognizes Pug templates as a file.

Shinigami92:
closes #105

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,60 @@
 
 [diff](https://github.com/prettier/plugin-pug/compare/1.5.1...1.6.0)
 
+## Big Change
+
+This plugin now supports alias options!  
+You can configure pug embedded in vue with them :tada:
+
+So instead of
+
+```js
+{
+  "singleQuote": true,
+  "overrides": [
+    {
+      "files": ["*.pug"],
+      "options": {
+        "singleQuote": false
+      }
+    }
+  ]
+}
+```
+
+you can now write
+
+```js
+{
+  "singleQuote": true,
+  "pugSingleQuote": false
+}
+```
+
+Aliases added for:
+
+- `pugPrintWidth`
+- `pugSingleQuote`
+- `pugTabWidth`
+- `pugUseTabs`
+- `pugBracketSpacing`
+- `pugSemi`
+
+Because prettier currently have something like a bug, you need to set e.g. `"pugSingleQuote": "true"` instead of `"pugSingleQuote": true`, if you want to set it to true  
+This needs to be done only for boolean `true` values
+
+If an alias is not set, the plugin will fallback to the non alias version  
+And in the special case of `singleQuote` option will take the inverse as before  
+So no breaking changes are introduced
+
+## Other Changes
+
 - Fix escape id for complex id values ([559ee6d])
 - Fix indentation of objects in multiline class attributes ([#110])
 
 [559ee6d]: https://github.com/prettier/plugin-pug/commit/559ee6d7f028e4eaba7768af70a9280d2190a2c2#diff-46bec2a2249807aa1515b246cc13ee93
 [#110]: https://github.com/prettier/plugin-pug/issues/110
+[#105]: https://github.com/prettier/plugin-pug/issues/105
 
 # 1.5.1
 
--- a/README.md
+++ b/README.md
@@ -135,6 +135,11 @@ yarn prettier --write "**/*.pug"
   Use tab for indentation
   Overrides `tabWidth`
 
+> All these six Prettier options have an alias e.g. `pugSingleQuote`  
+> You can force override pug formatting with them e.g. when using `pug` embedded in `vue` files
+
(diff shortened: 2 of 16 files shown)


SkyaTura:
Thanks for the fast response

Shinigami92:
@SkyaTura Can you provide me a sample `vue` file for `bracketSpacing` option that contains js and pug code?

SkyaTura:
My pain is actually with `singleQuote`:
  
```vue
<template lang="pug">
  // intended not to be single quoted
  ComponentBar(
    foo="bar"
    bar="baz"
  )
</template>

<script>
  export default {
    name: 'FooComponent' // intended to be single quote
  }
</script>
```

Shinigami92:
Good that I asked 😁
doesn't work as expected...
Need to work further

SkyaTura:
> OH WOW! I learned an interesting fact about prettier options :D

Could you mind to share what you've learn? I'm curious 😂

Shinigami92:
@SkyaTura Sure :)

Two things:

1. If you create aliases like `pug*` you also have to define the default values for this new option, otherwise it will not be taken at all
2. It's not really possible to have an option that has a default value of `null` or `undefined` when having a value like `int`  
    I'm currently struggling with prettier's default values, my `pug*` default values and incremental developing without breaking changes

SkyaTura:
Isn't possible to get access to the actual config declaration? Before prettier defaults been applied, so that you could check if there are `pug*` been set?

Or make `pug*` to be strings, so, if they are set you ignore the prettier's default?

_Just wondering here, I don't have any knowledge about prettier plugins yet_ 😰

Shinigami92:
Nope, nothing of this is possible :/

Shinigami92:
Worked on it almost exactly 3 hours 😁
Seems to work

I'm thinking to release a beta `1.6.0-beta.1`

And I also need to update changelog

Shinigami92:
@SkyaTura I released [1.6.0-issue-105.1](https://www.npmjs.com/package/@prettier/plugin-pug/v/1.6.0-issue-105.1) for you
Please test the hell out of it 😈

SkyaTura:
Sure I will

SkyaTura:
It's ~kinda~ working 😰 

I don't know if it's related to this project or I am doing something wrong, but I have this setup:

- A project where I define my team patterns as a eslint-config
- I use prettier plugin to load prettier in eslint
- I use the rules to override prettier's settings

**If** I run locally, on the module, as a test, it goes well, detect errors and fixes them as well.
**If** I run remotely, on a project, nothing happens.

The curious thing about it: using `@latest` works well, using the branch you've created, none of the rules are recognized.

_Bonus: this happens both on eslint and vim plugin_

So, basically, this is my files:

_module/prettier.js_
```js
module.exports = {
  extends: ['plugin:prettier/recommended'],
  plugins: ['prettier'],
  rules: {
    'prettier/prettier': [
      'error',
      {
        trailingComma: 'es5',
        printWidth: 80,
        tabWidth: 2,
        semi: false,
        singleQuote: true,
        endOfLine: 'lf',
        plugins: ['@prettier/plugin-pug'],
        pugSingleQuote: false,
        attributeSeparator: 'as-needed',
        closingBracketPosition: 'new-line',
        commentPreserveSpaces: 'trim-all',
      },
      {
        usePrettierrc: false,
      },
    ],
}
```

_project/.eslintrc.js_
```js
module.exports = {
  root: true,
  extends: [
    '@myOrg/eslint-config/prettier',
  ],
  ignorePatterns: ['node_modules', 'lib', 'dist', '.thalamus', '.nuxt'],
  rules: {},
}
```

If this isn't related, sorry :P

Shinigami92:
I need to fix a little thingy, but this will not affect you :)
Just the default assumption of non boolean default values needs to be changed

Cause currently it's not possible to define `{ "printWidth": xy, "pugPrintWidth": 80 }`, this would always take `xy` for everything

Shinigami92:
Dafuq?
Setting `"pugSingleQuote": true` results in `Invalid --pug-single-quote value. Expected null, false or "", but received true.` :eyes:  
I thing this is a bug in prettier  😲

And I also have a bug when using
```
{
  "singleQuote": true,
  "pugSingleQuote": false
}
```

This produces e.g. `div(:class="{ "pt-2": !dense }")` instead of `div(:class="{ 'pt-2': !dense }")`

Referencing https://github.com/prettier/prettier/issues/9204

PR #117, opened 2020-09-27
Pass on more prettier code-interpolation options

lehni:
Respect `bracketSpacing` and `arrowParens` options in code-interpolation


Shinigami92:
I need to investigate why the checks / tests are not running

Shinigami92 (changes requested):
Apparently you are missing the definitions in `/src/options/common.ts`

src/options/common.ts
```diff
@@ -96,6 +96,23 @@ export const PUG_BRACKET_SPACING_OPTION = {
 	]
 };
 
+export const PUG_ARROW_PARENS_OPTION = {
+	since: '1.6.2',
+	category: CATEGORY_PUG,
+	type: 'choice',
+	default: null,
+	choices: [
```
Shinigami92, line 104: You should copy the descriptions defined by prettier itself: https://github.com/prettier/prettier/blob/6acf5370f7154ac66cd0db270c71b60e5e136eca/src/language-js/options.js#L9-L28
_all 3_

Also swap both objects, I don't know if this will have an effect on CLI suggestions

src/printer.ts
```diff
@@ -827,8 +831,7 @@ export class PugPrinter {
 			val = format(val, {
 				parser: 'babel',
 				...this.codeInterpolationOptions,
-				semi: useSemi,
-				endOfLine: 'lf'
```
Shinigami92, line 831: Did you remove `endOfLine: 'lf'` because it's already defined in `codeInterpolationOptions`?

src/printer.ts
```diff
@@ -79,6 +79,8 @@ export interface PugPrinterOptions {
 	readonly pugUseTabs: boolean;
 	readonly bracketSpacing: boolean;
 	readonly pugBracketSpacing: boolean;
+	readonly arrowParens: 'avoid' | 'always';
```
Shinigami92, line 82: Maybe we should extract a `export type ArrowParens = 'avoid' | 'always';` to `src/options/common.ts`

src/options/index.ts
```diff
@@ -22,12 +23,13 @@ import {
...
 
 export interface PugParserOptions
-	extends Pick<ParserOptions, 'printWidth' | 'singleQuote' | 'tabWidth' | 'useTabs' | 'bracketSpacing' | 'semi'> {
+	extends Pick<ParserOptions, 'printWidth' | 'singleQuote' | 'tabWidth' | 'useTabs' | 'bracketSpacing' | 'arrowParens' | 'semi'> {
 	pugPrintWidth: number;
 	pugSingleQuote: boolean | null;
 	pugTabWidth: number;
 	pugUseTabs: boolean | null;
 	pugBracketSpacing: boolean | null;
+	pugArrowParens: 'avoid' | 'always' | null;
```
Shinigami92, line 32: Use new type `pugArrowParens: ArrowParens | null` here

Shinigami92 (changes requested):

src/options/common.ts
lehni: Ah, i was looking for this but couldn't find it. Perfect!

lehni: What do you mean with _all 3_?

src/printer.ts
lehni: Yes indeed, it seemed redundant to mix it in again here?

src/printer.ts
lehni: Agreed

src/options/common.ts
lehni: Ah, `null` needs to be included also, go tit.

lehni:
I think I addressed all raised points now.

Shinigami92:
The checks are running :tada:

src/printer.ts
Shinigami92: Yeah, I think the only real reason why this was there was a theoretical reason, cause if you don't pass `endOfLine: 'lf'` here, the formatter breaks.
And if someone want to add `endOfLine / pugEndOfLine` this will not be remembered and the suddenly break :bomb:

lehni:
All green! (Not the great [musician](https://www.theguardian.com/music/musicblog/2016/apr/20/al-green-10-of-the-best), but also good ;) )

src/printer.ts
lehni: Ok I'll put it back!

Shinigami92:
Small trick for you when working on PRs in other repos



You can re-request a review :slightly_smiling_face:

Shinigami92 (approved):

PR #119, opened 2020-09-29
Implement pugWrapAttributesThreshold option

### LINKED ISSUE #118: Option to control threshold of amount of attributes to be put on separate lines
## Request / Idea

It would be nice if there was an option with which I could set a threshold of the amount of attributes that should always be put on separate lines, regardless of wether they fit into `printWidth` chars or not, .e.g.

```js
attributesOnNewLineThreshold: 2
```

## Input

```pug
.container-with-one-attribute(attribute1='foo')

.container-with-two-attributes(attribute1='foo' attribute2='bar')
```

## Expected Output

```pug
.container-with-one-attribute(attribute1='foo')

.container-with-two-attributes(
  attribute1='foo'
  attribute2='bar'
)
```

Would you consider the addition of such an option? I am happy to work on it. Not sure what it should be called though…

lehni:
Closes #118


Shinigami92:
1. Please use draft PRs
2. Please request me as a reviewer if you are done

:slightly_smiling_face: 

If you don't have these options, please let me know and I need to check my repo settings

src/options/converge.ts
```diff
@@ -22,6 +22,7 @@ export function convergeOptions(options: ParserOptions & PugParserOptions): PugP
 		closingBracketPosition: options.pugClosingBracketPosition ?? options.closingBracketPosition,
 		commentPreserveSpaces: options.pugCommentPreserveSpaces ?? options.commentPreserveSpaces,
 		pugSortAttributesBeginning: options.pugSortAttributesBeginning,
-		pugSortAttributesEnd: options.pugSortAttributesEnd
+		pugSortAttributesEnd: options.pugSortAttributesEnd,
+		pugMaxAttributesOnFirsLine: options.pugMaxAttributesOnFirsLine
```
Shinigami92, line 26: You have a typo in your copy-paste 😛

Shinigami92 (changes requested):

Shinigami92 (changes requested):
I don't like the option name `pugMaxAttributesOnFirstLine`, because it confuses me and indicates the first line of the `document` and not the `tag`

So maybe `pugMaxAttributesOnOneLine`? Or `pugNumAttributesWithoutWrap` or something similar
Cause why `max`? Taken a maximum possibly indicates that less is _what_?

lehni:
Yes I'm also not convinced with the name yet. I thought about  `...OnOneLine` also, but then you'd expect that to apply to the other lines as well, allowing more than 1 attribute on the 2nd line, the 3rd of an element, etc.

`max` because there are other causes of a wrap, mainly `printWidth`. So this is the maximum of attributes that is ever allowed to be on the first line of an element, unless something else forces them to wrap already.

With `pugNumAttributesWithoutWrap` I would expect this to always apply, regardless of `printWidth`

lehni:
How about `pugMaxAttributesPerLine`? It's simple, and even though it wouldn't apply to the other lines it's still true: If wrapping is enforced, every line only has one attribute anyway.

src/options/converge.ts
lehni: Oh hahaha. It was late!

lehni: It was actually a sloppy search'n'replace 🤕

lehni:
Oh damn, `pugMaxAttributesPerLine` also doesn't work, because you can set it to `0`, forcing a wrap right away:

```pug
.element(foo="bar")
```
->
```pug
.element(
  foo="bar"
)
```

lehni:
Some other possibilities:

* `pugWrapAttributesThreshold`
* `pugWrapAttributes` 

Perhaps `pugWrapAttributes` is good. It could be used like this:

```js
pugWrapAttributes: true // always wrap, even with only one attribute
pugWrapAttributes: 2 // only wrap when there are two or more attributes, or when they don't fit on `printWidth`
```

Shinigami92:
`pugMaxAttributesWithoutWrap` would be an okay-ish option
I liked `pugMaxAttributesPerLine`, but I see your point :thinking: 

`pugWrapAttributes` with `boolean` and `int` is not accepted by prettier option parser :-1: 
```yaml
pugWrapAttributes: true
pugWrapAttributes: 2
```

lehni:
Ok, new idea, my favourite so far:

```js
pugWrapAttributes: 'as-needed' // the default, as is now
pugWrapAttributes: 'always' // same as `0` above
pugWrapAttributes: 'two-or-more' // same as `1` above 
pugWrapAttributes: 'three-or-more' // same as `2` above 
```

I don't think you need more than these four options, really!

The naming of the last two options could change. Some alternatives:

```js
'above-one', 'above-two'
'>1', '>2'
```

Shinigami92:
But then the next is coming and having a `printWidth` of `160` and want 'four-or-more'?
I don't know if this is good :thinking: 

_what if working with a 21:9 monitor_ :smile:

lehni:
Yeah, of course you're right…. ARGH.

But why not add two options then?

```js
pugWrapAttributes: 'as-needed' // the default, as is now
pugWrapAttributes: 'always' // same as `0` above
pugWrapAttributes: 'above-threshold' // this requires the separate property to be set:
 
pugWrapAttributesThreshold: 2
```

Shinigami92:
We could use two options, but I personally don't like such option splitting where one option needs to set another option :man_shrugging:  
Especially if it could be configured in a way with `-1, 0, 1, 2, 3 .. Infinity`

lehni:
But the naming is nicer with the split? Hmmmm.

Shinigami92:
I don't thing that the naming should be that nice that it was split for that
Then I think some of the other (one-option) alternatives are better

lehni:
Let's choose one of these?

`pugMaxAttributesWithoutWrap`
`pugMaxUnwrappedAttributes`
`pugWrapAttributesThreshold`

Shinigami92:
`pugMaxAttributesWithoutWrap` or `pugWrapAttributesThreshold`
Both of them needs documentation in readme anyway

Shinigami92:
- `pugMaxAttributesWithoutWrap`
I don like the max and _negative_ without

Ok lets take: `pugWrapAttributesThreshold`

lehni:
I agree, good choice!

tests/options/pugWrapAttributesThreshold/0/0.test.ts
```diff
@@ -0,0 +1,21 @@
+import { readFileSync } from 'fs';
+import { resolve } from 'path';
+import { format } from 'prettier';
+import { plugin } from './../../../../src/index';
+
+describe('Options', () => {
+	describe('pugWrapAttributesThreshold', () => {
+		test('should never allow an attribute on the first line of an element', () => {
```
Shinigami92, line 8:
```suggestion
		test('should never allow an attribute without being wrapped', () => {
```

tests/options/pugWrapAttributesThreshold/1/1.test.ts
```diff
@@ -0,0 +1,21 @@
+import { readFileSync } from 'fs';
+import { resolve } from 'path';
+import { format } from 'prettier';
+import { plugin } from './../../../../src/index';
+
+describe('Options', () => {
+	describe('pugWrapAttributesThreshold', () => {
+		test('should not allow more than one attribute on the first line of an element', () => {
```
Shinigami92, line 8:
```suggestion
		test('should not allow more than one attribute as one-liner', () => {
```

tests/options/pugWrapAttributesThreshold/2/2.test.ts
```diff
@@ -0,0 +1,21 @@
+import { readFileSync } from 'fs';
+import { resolve } from 'path';
+import { format } from 'prettier';
+import { plugin } from './../../../../src/index';
+
+describe('Options', () => {
+	describe('pugWrapAttributesThreshold', () => {
+		test('should not allow more than two attributes on the first line of an element', () => {
```
Shinigami92, line 8:
```suggestion
		test('should not allow more than two attributes as one-liner', () => {
```

tests/options/pugWrapAttributesThreshold/default/default.test.ts
```diff
@@ -0,0 +1,21 @@
+import { readFileSync } from 'fs';
+import { resolve } from 'path';
+import { format } from 'prettier';
+import { plugin } from './../../../../src/index';
+
+describe('Options', () => {
+	describe('pugWrapAttributesThreshold', () => {
+		test('should not restrict the amount of attributes on the first line of an element', () => {
```
Shinigami92, line 8:
```suggestion
		test('should not restrict the amount of attributes as one-liner', () => {
```

Shinigami92 (changes requested):

Shinigami92 (approved):

lehni:
Thanks for the merge @Shinigami92! I just noticed that I forgot to add docs about it to the README, will to this now. Does this merrit a 1.8.0 release? That's what it says here: https://github.com/prettier/plugin-pug/blob/0fa265ba208efb2725104613fdc1a3f6297f4c68/src/options/wrap-attributes-threshold.ts#L4
😉

Shinigami92:
Yes this will be in a 1.8.0 release, I already thought that I will write Changelog and package.json

If you want to write something in readme, yes pls do

Shinigami92:
But our other PR and another PR will also possibly in 1.8.0
So I will have 4 new PRs in one release

Shinigami92:
https://twitter.com/pr0grammerhum0r/status/1311395456153464834?s=21

lehni:
@Shinigami92 I was playing around with this new option yesterday and our rather large code-base where the pug templates were hand-crafted so far. One thing I noticed is that when the properties are "normal" HTML properties, we mostly put them on one line, but as soon as there is some logic or code involved, we've wrapped them for increased visibility and easier parsing of the code. I should have thought about this before writing the PR perhaps.

I was toying with an additional feature, which is very easy to implement, and takes hints from the attribute sorting options. Would you consider adding this?

```js
pugWrapAttributesPattern: "^(v-|:|@)"
```

It can provide a regular expression that triggers wrapping of all attributes as soon one attribute tests positively against it.

Shinigami92:
IMO no

We should write into `README.md` that these attributes needs precedence over other attributes and therefore should be configured in `pugSortAttributesBeginning`  

Think about vue 3 v-if / v-for and vue 2 v-for / v-if
And then there is angular
And then there are other unknown underdog frameworks
...

Shinigami92:
Or maybe I missunderstood you? cause it's not sorting related?! 😕 

Can you provide an example input / expected output / actual output?

Maybe open a new issue for that? So we are not talking about it in a merged PR

lehni:
No, it's not sort related. Here's a real world example:

Some SVG pug for icons:

```pug
      circle(cx="12" cy="12" r="11")
      line(x1="12" y1="6" x2="12" y2="13")
      circle(cx="12" cy="17" r="0.5")
```

VS some Vue pug with logic:

```pug
div(
  v-if="isList"
)
  delayed-unhide-html(
    v-for="(node, index) in list"
    :key="index"
    :markup="node"
    :options="options"
  )
``` 

It's been our convention to wrap these, even when there's only one attribute, because it makes the logic in the template more obvious. But then we wrap everything, and the nodes that have only normal HTML attributes and often many of them become very long.

Hence the idea to add this `pugWrapAttributesPattern` option, which which I could trigger wrapping for _logic_  attributes.

`/^(v-|:|@)/` : Any attribute that starts with `v-`, `:`, or `@`

Shinigami92:
Looks very strange to me, but create an Issue with your example (for tracking and referencing later in changelog)
and open a PR

But maybe your issue is related to not wrap attributes in `svg` tag? 🤷‍♂️

PR #120, opened 2020-09-30
Add middle attributes sorting

SkyaTura:
## TODO

- [ ] Explain this PR

#### Original post
<details>
## Reason for this PR

Every time this code is saved it gets a different result:

```vue
<template lang="pug">
Foo(c b a d i f g h e l n)
</template>
```

## Expected result

```vue
<template lang="pug">
Foo(a b c d e f g h i l n)
</template>
```
<details>

### DIFF
--- a/src/options/attribute-sorting/index.ts
+++ b/src/options/attribute-sorting/index.ts
@@ -24,7 +24,7 @@ export const PUG_SORT_ATTRIBUTES_OPTION = {
 	category: CATEGORY_PUG,
 	type: 'choice',
 	default: 'as-is',
-	description: 'Change how the attributes between Beggining and End should be sorted',
+	description: 'Change how the attributes between _beginning_ and _end_ should be sorted.',
 	choices: [
 		{ value: 'asc', description: 'Sort middle attributes ascending.' },
 		{ value: 'desc', description: 'Sort middle attributes descending.' },
--- a/src/options/attribute-sorting/index.ts
+++ b/src/options/attribute-sorting/index.ts
@@ -18,3 +18,19 @@ export const PUG_SORT_ATTRIBUTES_END_OPTION = {
 	...pugSortAttributesOption,
 	description: 'Define a list of patterns for attributes that are sorted at the end.'
 };
+
+export const PUG_SORT_ATTRIBUTES_OPTION = {
+	since: '1.8.0',
+	category: CATEGORY_PUG,
+	type: 'choice',
+	array: false,
+	default: 'as-is',
+	description: 'Change how the attributes between Beggining and End should be sorted',
+	choices: [
+		{ value: 'asc', description: 'Sort middle attributes ascending.' },
+		{ value: 'desc', description: 'Sort middle attributes descending.' },
+		{ value: 'as-is', description: 'Middle attributes are leave untouched.' }
+	]
+};
+
+export type SortAttributes = 'asc' | 'desc' | 'as-is';
--- a/src/options/attribute-sorting/index.ts
+++ b/src/options/attribute-sorting/index.ts
@@ -23,7 +23,6 @@ export const PUG_SORT_ATTRIBUTES_OPTION = {
 	since: '1.8.0',
 	category: CATEGORY_PUG,
 	type: 'choice',
-	array: false,
 	default: 'as-is',
 	description: 'Change how the attributes between Beggining and End should be sorted',
 	choices: [
--- a/src/options/attribute-sorting/utils.ts
+++ b/src/options/attribute-sorting/utils.ts
@@ -2,13 +2,14 @@ import { AttributeToken } from 'pug-lexer';
 import { SortAttributes } from './index';
 
 type CompareResult = -1 | 0 | 1;
+type CompareFunction = (a: any, b: any) => CompareResult;
 
 export function compareAttributeToken(
 	a: AttributeToken,
 	b: AttributeToken,
-	sortAttributesEnd: string[],
+	sortAttributes: SortAttributes,
 	sortAttributesBeginning: string[],
-	sortAttributes: SortAttributes
+	sortAttributesEnd: string[]
 ): CompareResult {
 	const sortPatternsBeginning: RegExp[] = sortAttributesBeginning.map((sort) => new RegExp(sort)).reverse();
 	const sortPatternsEnd: RegExp[] = sortAttributesEnd.map((sort) => new RegExp(sort));
@@ -41,9 +42,20 @@ export function compareAttributeToken(
 	return 0;
 }
 
-export function partialSort<T>(arr: T[], start: number, end: number, compareFn?: (a: T, b: T) => number): T[] {
+export function stableSort<T>(array: T[], compare: CompareFunction): any[] {
+	const entries = array.map((value, index): [T, number] => [value, index]);
+	entries.sort((a, b) => {
+		const order = compare(a[0], b[0]);
+		// When order is 0, sort by index to make the sort stable:
+		return order != 0 ? order : a[1] - b[1];
+	});
+	return entries.map(([value]) => value);
+}
+
+export function partialSort<T>(arr: T[], start: number, end: number, compareFn: CompareFunction): T[] {
 	const preSort: T[] = arr.slice(0, start);
 	const postSort: T[] = arr.slice(end);
-	const sorted: T[] = arr.slice(start, end).sort(compareFn);
+	const attributes: T[] = arr.slice(start, end);
+	const sorted: T[] = stableSort(attributes, compareFn);
 	return [...preSort, ...sorted, ...postSort];
 }
--- a/src/options/attribute-sorting/utils.ts
+++ b/src/options/attribute-sorting/utils.ts
@@ -1,4 +1,4 @@
-import { AttributeToken } from 'pug-lexer';
+import {AttributeToken} from 'pug-lexer';
 
 type CompareResult = -1 | 0 | 1;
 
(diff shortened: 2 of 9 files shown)


SkyaTura:
I've accidentally formatted things that I shouldn't on a formatter library :P

Going to fix it right now!

Shinigami92:
1. Please use draft PRs
2. Please request me as a reviewer if you are done

:slightly_smiling_face:

Shinigami92:
Please add a prettier config to the issue descriptions

Currently I would expect that nothing would get sorted cause the default would be taken

Also please note that the sorting isn't sorting in an alphabetical order in any way

src/options/attribute-sorting/utils.ts
```diff
@@ -22,47 +22,38 @@ function compareByIndex(leftIndex: number, rightIndex: number): CompareResult {
...
-			}
-			if (!bFound && pattern.test(bName)) {
-				bIndex = index;
-				bFound = true;
-			}
-			if (aFound && bFound) {
-				break;
-			}
-		}
+	if (beginning > 0) return -1;
```
Shinigami92, line 38: Oh I need to improve my eslint...

I really don't like these non curly braces ^^'

src/options/attribute-sorting/utils.ts
```diff
@@ -22,47 +22,38 @@ function compareByIndex(leftIndex: number, rightIndex: number): CompareResult {
...
 }
 
 export function partialSort<T>(arr: T[], start: number, end: number, compareFn?: (a: T, b: T) => number): T[] {
-	const preSorted: T[] = arr.slice(0, start);
-	const postSorted: T[] = arr.slice(end);
-	const sorted: T[] = arr.slice(start, end).sort(compareFn);
-	arr.length = 0;
-	arr.push(...preSorted.concat(sorted).concat(postSorted));
-	return arr;
+	const preSort: T[] = [...arr].slice(0, start);
```
Shinigami92, line 55: Please explain why these three `[...arr]` are used, I want to understand :slightly_smiling_face:

tests/options/sortAttributes/sort-attributes.test.ts
```diff
@@ -38,19 +38,62 @@ describe('Options', () => {
...
+		test('compare 3', () => {
+			const pugSortAttributesBeginning = ['^x$', '^y$', '^z$'];
+			const pugSortAttributesEnd = ['v-for', ':key', 'src', 'alt', '@click', ':disabled'];
+			const expected: ReadonlyArray<string> = [
+				'x',
+				'y',
+				'z',
+				'a',
+				'b',
+				'c',
```
Shinigami92, line 81: Why would `a` sorted before `c` if you didn't configured it in any way in `beginning` or `end`?
I won't expect this `expected`

I would expect this: `'x', 'y', 'z', 'c', 'a', 'b', 'v-for', ':key', 'src', 'alt', '@click', ':disabled'`

Shinigami92 (changes requested):

src/options/attribute-sorting/utils.ts
SkyaTura: I'll add those later ;)

src/options/attribute-sorting/utils.ts
SkyaTura: I reverted this change, it was useless.

tests/options/sortAttributes/sort-attributes.test.ts
SkyaTura: I was considering that middle attributes should be sorted alphabetically, which I was, indeed, wrong.

_this is what happens when you try to mess with someone's project while sleepy, sorry_

SkyaTura:
So, I'm thinking and experimenting things for long hours now, but I can't find an answer.

What have I done since my first changes:

- In fact, this formatter had nothing to do with "middle attributes", so I've added a new option `pugSortAttributes = 'as-is' | 'asc' | 'desc'`. This could be reverted if you think it shouldn't exist.
- Added more test cases, with different middle attribute sorting options.

What I cannot understand is, why on earth `compare 3` and `compare 6` are failing.

src/options/attribute-sorting/index.ts
```diff
@@ -18,3 +18,19 @@ export const PUG_SORT_ATTRIBUTES_END_OPTION = {
 	...pugSortAttributesOption,
 	description: 'Define a list of patterns for attributes that are sorted at the end.'
 };
+
+export const PUG_SORT_ATTRIBUTES_OPTION = {
+	since: '1.8.0',
+	category: CATEGORY_PUG,
+	type: 'choice',
+	array: false,
```
Shinigami92, line 26: You don't need to pass `array: false`, cause it's the default

Shinigami92 (changes requested):

Shinigami92:
Can you rename the PR title? :)

Shinigami92:
@SkyaTura Looking at the tests, it seems that your sort is not stable :thinking:

SkyaTura:
Yes, but I don't know why. My knowledge is limited here, because I though it should work correctly.

I'll study more about the sort method to understand what is going on, over this weekend, _( the last couple days have been busy for me :x )_

lehni:
Could it be the cases where the compare function returns `0` that leads to unstable sort?

Node's `sort()` wasn't stable up to about Node 11 or so.

Some bits about this:

https://bugs.chromium.org/p/v8/issues/detail?id=90
https://github.com/nodejs/node/issues/29446#issuecomment-528164143
https://medium.com/@fsufitch/is-javascript-array-sort-stable-46b90822543f

I would try a stable `sort()` method next, e.g.:

```js
function stableSort(
  array,
  compare = (a, b) => a < b ? -1 : a > b ? 1 : 0
 ) {
  const entries = array.map((value, index) => [value, index]);
  entries.sort((a, b) => {
    const order = compare(a[0], b[0]);
    // When order is 0, sort by index to make the sort stable:
    return order != 0 ? order : a[1] - b[1];
  });
  return entries.map(([value]) => value);
}
```

Note that this doesn't modify `array`, it returns a copy.

SkyaTura:
Finally!!! My logic wasn't wrong after all 😅 

Thank you a lot @lehni . This stableSort worked as a chamr.

Shinigami92:
@SkyaTura @lehni Check out my newly added contributing.me 😃
Maybe it could help 🙂

lehni:
@SkyaTura happy to hear that this helped! I was struggling with this situation a while back on Node also, it can be so confusing.

src/options/attribute-sorting/utils.ts
```diff
@@ -1,68 +1,77 @@
...
+	const end = aEndIndex - bEndIndex;
+	if (end > 0) {
+		return 1;
+	}
+	if (end < 0) {
+		return -1;
+	}
 
-		result = compareByIndex(aIndex, bIndex);
+	if (sortAttributes === 'asc') {
```
Shinigami92, line 42: Can you convert this to a switch case? Then it could be extend better in future

src/options/attribute-sorting/utils.ts
```diff
@@ -1,68 +1,77 @@
...
-	return result;
+	return 0;
+}
+
+export function stableSort<T>(array: T[], compare: CompareFunction): any[] {
+	const entries = array.map((value, index): [T, number] => [value, index]);
+	entries.sort((a, b) => {
+		const order = compare(a[0], b[0]);
+		// When order is 0, sort by index to make the sort stable:
+		return order != 0 ? order : a[1] - b[1];
```
Shinigami92, line 66: Always prefer `!==`

src/options/attribute-sorting/utils.ts
```diff
@@ -1,68 +1,77 @@
...
+		if (aName < bName) {
+			return 1;
+		}
 	}
 
-	return result;
+	return 0;
+}
+
+export function stableSort<T>(array: T[], compare: CompareFunction): any[] {
```
Shinigami92, line 61: Not quite sure, but isn't it `: T[]` as return type?

tests/options/sortAttributes/sort-attributes.test.ts
```diff
@@ -38,19 +39,181 @@ describe('Options', () => {
...
+			const expected: ReadonlyArray<string> = 'aedcfghilnb'.split('');
+			const code: string[] = 'aedcfghilnb'.split('');
+			const actual: string[] = stableSort(code, (a, b) =>
+				compareAttributeToken(
+					createAttributeToken(a),
+					createAttributeToken(b),
+					pugSortAttributes,
+					pugSortAttributesBeginning,
+					pugSortAttributesEnd
+				)
```
Shinigami92, line 216: Can you add a test using beginning and end but don't pass `pugSortAttributes` so it takes the default one?

tests/options/sortAttributes/sort-attributes.test.ts
```diff
@@ -38,19 +39,181 @@ describe('Options', () => {
...
+			);
+
+			expect(actual).toStrictEqual(expected);
+		});
+		test('compare 5', () => {
```
Shinigami92, line 154: I know it's hard, but I thing we should try to find names for all these `compare x` tests
Now there are 7 and we cannot read from them what they are doing ^^

Shinigami92 (changes requested):

src/options/attribute-sorting/utils.ts
SkyaTura: To be honest, I'm new to TypeScript, I didn't knew how to properly type the method @lehni suggested.

Could you help me with this one?

Shinigami92: Ah okay, I can checkout your branch at evening and then suggest some types

tests/options/sortAttributes/sort-attributes.test.ts
Shinigami92: @SkyaTura Nice one, couldn't make it better ✨

Shinigami92 (approved):

PR #121, opened 2020-09-30
Fix counting of current line-length

lehni:
I noticed that the current line-length wasn't counted correctly depending on the setting of attributeSeparator.  
This PR fixes that and adds tests for the various settings.

### DIFF
--- a/src/printer.ts
+++ b/src/printer.ts
@@ -235,10 +235,9 @@ export class PugPrinter {
 	}
 
 	private tokenNeedsSeparator(token: AttributeToken): boolean {
-		return (
-			!this.neverUseAttributeSeparator &&
-			(this.alwaysUseAttributeSeparator || /^(\(|\[|:).*/.test(token.name))
-		);
+		return this.neverUseAttributeSeparator
+			? false
+			: this.alwaysUseAttributeSeparator || /^(\(|\[|:).*/.test(token.name);
 	}
 
 	private formatDelegatePrettier(
@@ -451,9 +450,8 @@ export class PugPrinter {
 						this.currentLineLength += tempToken.name.length;
 						if (hasNonPrefixAttributes) {
 							// This isn't the first non-prefix attribute, add a space and separator
+							this.currentLineLength += 1;
 							if (this.tokenNeedsSeparator(tempToken)) {
-								this.currentLineLength += 2;
-							} else {
 								this.currentLineLength += 1;
 							}
 						}
--- a/src/printer.ts
+++ b/src/printer.ts
@@ -100,14 +100,10 @@ export class PugPrinter {
 	private result: string = '';
 
 	private currentIndex: number = 0;
+	private currentLineLength: number = 0;
 
 	private readonly indentString: string;
 	private indentLevel: number = 0;
-
-	/**
-	 * Counting the line length starts at 0
-	 */
-	private currentLineLength: number = 0;
 
 	private readonly quotes: "'" | '"';
 	private readonly otherQuotes: "'" | '"';
@@ -420,21 +416,24 @@ export class PugPrinter {
 			this.possibleClassPosition = this.result.length;
 			result = '(';
 			logger.debug(this.currentLineLength);
-			this.currentLineLength += 1;
 			let tempToken: AttributeToken | EndAttributesToken = this.nextToken;
 			let tempIndex: number = this.currentIndex + 1;
-			let hasPrefixAttribute: boolean = false;
-			let hasNonPrefixAttributes: boolean = false;
-			let numAttributes: number = 0;
+			// In pug, tags can have two kind of attributes: normal attributes that appear between parantheses,
+			// and literals for ids and classes, prefixing the paranthesis, e.g.: `#id.class(attribute="value")`
+			// https://pugjs.org/language/attributes.html#class-literal
+			// https://pugjs.org/language/attributes.html#id-literal
+			// In the stream of attribute tokens, distinguish those that can be converted to literals,
+			// and count those that cannot (normal attributes) to determine the resulting line length correctly.
+			let hasLiteralAttributes: boolean = false;
+			let numNormalAttributes: number = 0;
 			while (tempToken.type === 'attribute') {
-				numAttributes++;
 				if (!this.wrapAttributes && this.wrapAttributesPattern?.test(tempToken.name)) {
 					this.wrapAttributes = true;
 				}
 				switch (tempToken.name) {
 					case 'class':
 					case 'id': {
-						hasPrefixAttribute = true;
+						hasLiteralAttributes = true;
 						const val: string = tempToken.val.toString();
 						if (isQuoted(val)) {
 							this.currentLineLength -= 2;
@@ -448,8 +447,9 @@ export class PugPrinter {
 					}
 					default: {
 						this.currentLineLength += tempToken.name.length;
-						if (hasNonPrefixAttributes) {
-							// This isn't the first non-prefix attribute, add a space and separator
+						if (numNormalAttributes > 0) {
+							// This isn't the first normal attribute that will appear between parantheses,
+							// add space and separator
 							this.currentLineLength += 1;
 							if (this.tokenNeedsSeparator(tempToken)) {
 								this.currentLineLength += 1;
@@ -464,32 +464,29 @@ export class PugPrinter {
 							this.currentLineLength += 1 + val.length;
 							logger.debug({ tokenVal: val, length: val.length }, this.currentLineLength);
 						}
-						hasNonPrefixAttributes = true;
+						numNormalAttributes++;
 						break;
 					}
 				}
 				tempToken = this.tokens[++tempIndex] as AttributeToken | EndAttributesToken;
 			}
 			logger.debug('after token', this.currentLineLength);
-			if (hasPrefixAttribute) {
-				// Remove div
+			if (hasLiteralAttributes) {
+				// Remove div as it will be replaced with the literal for id and/or class
 				if (this.previousToken?.type === 'tag' && this.previousToken.val === 'div') {
 					this.currentLineLength -= 3;
 				}
 			}
-			if (hasNonPrefixAttributes) {
-				// Add trailing brace
-				this.currentLineLength += 1;
-			} else {
-				// Remove leading brace
-				this.currentLineLength -= 1;
+			if (numNormalAttributes > 0) {
+				// Add leading and trailing parantheses
+				this.currentLineLength += 2;
 			}
 			logger.debug(this.currentLineLength);
 			if (
 				!this.wrapAttributes &&
 				(this.currentLineLength > this.options.pugPrintWidth ||
 					(this.options.pugWrapAttributesThreshold >= 0 &&
-						numAttributes > this.options.pugWrapAttributesThreshold))
+						numNormalAttributes > this.options.pugWrapAttributesThreshold))
 			) {
 				this.wrapAttributes = true;
 			}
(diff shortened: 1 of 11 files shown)


src/printer.ts
```diff
@@ -101,9 +101,9 @@ export class PugPrinter {
 	private indentLevel: number = 0;
 
 	/**
-	 * The line length starts by 1, it's not an zero based index
+	 * Counting the line length starts at 0
 	 */
-	private currentLineLength = 1;
+	private currentLineLength = 0;
```
Shinigami92, line 106: It's indeed index 1 :warning:  
Please try to start with an `currentLine*` based on `1`

Shinigami92 (changes requested):

src/printer.ts
lehni: If you turn on the word wrap column in VSCode and set it to 80 and look at the tests that I added, you will see that you get the expected results with 0, not 1…

image

Shinigami92: Ok need to check why I had used this index 1 base
Cause there also was another reason, but I don't know anymore

Need to start working _Home Office_, so I will only check time to time here :wave: 
Thanks for all our work :slightly_smiling_face:

lehni: Of course, no worries! And thank you for the great work on this plugin, it's exactly what we need!

lehni: Maybe you were using the `as-needed` option where the counting is off currently?

Shinigami92: Could be, need to look at the code again
It's a long time ago I read this part of code ^^'

lehni: I know the feeling 😆

tests/options/attributeSeparator/always/always.test.ts
```diff
@@ -11,6 +11,7 @@ describe('Options', () => {
 			const actual: string = format(code, {
 				parser: 'pug' as any,
 				plugins: [plugin],
+				printWidth: 80,
```
Shinigami92, line 14: Why is this needed? I mean `80` is the default
If it is intended cause any reason, you could add a comment, so we also now in 100+ years (if pug is also then a thing)

lehni: I added it explicitly because the test runs against that `printWidth`. Kinda the same reason why you added `endOfLine: 'lf'` ;)

Happy to remove it though

Shinigami92: We should add a comment, and yes I will also add a comment on the `endOfLine: 'lf'` thingy later (when my working day has ended)

Shinigami92:
I really like the `tokenNeedsSeparator` member function, maybe we should extract it as a _exported_ function to `src/utils/common.ts`. Then we could also write tests specifically for it :thinking: 

For an in depth review, I need more (free-)time :slightly_smiling_face:

lehni:
Thanks! The function relies on state on the class currently which is read from the options (`this.neverUseAttributeSeparator` and  `this.alwaysUseAttributeSeparator`), so doing this would be a bit tricky! Also, I'm not sure we need to test the function in isolation, since we're already testing its effects through the many tests relating to `attributeSeparator`.

Shinigami92:
Sorry for the merge conflicts, cannot resolve them on my side :slightly_frowning_face:  
That will change when you later create branches in this repo directly :slightly_smiling_face:

lehni:
No worries! Will resolve later!

lehni:
@Shinigami92 I've rebased the changes in this branch on top of master, resolving all the conflicts, and force-pushed a new version. I hope that's fine!

Shinigami92:
@lehni I hope I'm not to deep into _"how I organize PRs"_, but is this feature merge ready?
If now more work is **planned** on your side, you should click the `Ready for review` Button and also the `Re-request review` Icon :slightly_smiling_face:  
(Applies to all PRs)

src/printer.ts
```diff
@@ -230,6 +230,13 @@ export class PugPrinter {
 		return !!token && possibilities.includes(token.type) !== invert;
 	}
 
+	private tokenNeedsSeparator(token: AttributeToken): boolean {
+		return (
+			!this.neverUseAttributeSeparator &&
+			(this.alwaysUseAttributeSeparator || /^(\(|\[|:).*/.test(token.name))
```
Shinigami92, line 236: You should run `yarn format` if you are finished with your work

This will become 
```ts
return (
	!this.neverUseAttributeSeparator && (this.alwaysUseAttributeSeparator || /^(\(|\[|:).*/.test(token.name))
);
```

Sadly this is not well readable

So maybe rewrite it to
```ts
if (this.neverUseAttributeSeparator) {
  return false;
}
return this.alwaysUseAttributeSeparator || /^(\(|\[|:).*/.test(token.name));
```

This is longer but increases readability, I think

src/printer.ts
```diff
@@ -413,8 +420,8 @@ export class PugPrinter {
 			this.currentLineLength += 1;
 			let tempToken: AttributeToken | EndAttributesToken = this.nextToken;
 			let tempIndex: number = this.currentIndex + 1;
-			let nonPrefixAttributes: number = 0;
 			let hasPrefixAttribute: boolean = false;
+			let hasNonPrefixAttributes: boolean = false;
```
Shinigami92, line 424: I even didn't know myself what `hasPrefixAttribute` represents :sweat_smile: 
But isn't `hasNonPrefixAttributes` the opposite of `hasPrefixAttribute`
Maybe some inline comments could help here

src/printer.ts
```diff
@@ -434,8 +441,15 @@ export class PugPrinter {
...
 					}
 					default: {
-						nonPrefixAttributes += 1;
 						this.currentLineLength += tempToken.name.length;
+						if (hasNonPrefixAttributes) {
+							// This isn't the first non-prefix attribute, add a space and separator
+							if (this.tokenNeedsSeparator(tempToken)) {
+								this.currentLineLength += 2;
+							} else {
+								this.currentLineLength += 1;
```
Shinigami92, line 450: Could this be "simplified" to

```ts
this.currentLineLength += 1;
if (this.tokenNeedsSeparator(tempToken)) {
  this.currentLineLength += 1;
}
```

Or does this have a technical reason?

Shinigami92 (changes requested):

src/printer.ts
lehni: Yes I actually had it like this first. But the part in my brain that used to write assembler code for Intel Pentium wanted to reduce the amount of CPU cycles per path, hence doing the above. 🤣 

Happy to change.

src/printer.ts
lehni: Oh I thought `release.sh` runs `yarn format` also. My bad!

lehni: Also, maybe `yarn lint` should complain about things that `yarn format` would do differently? BTW, why not just use prettier on the code base? That would reduce a lot of these discussions as there's only one solution 😉

lehni: Could also be:

```js
		return this.neverUseAttributeSeparator
			? false
			: this.alwaysUseAttributeSeparator || /^(\(|\[|:).*/.test(token.name);
```

If you like ternaries (I do ;) )

Shinigami92: The one thing with `prettier` is that I format on **save** :laughing:

src/printer.ts
lehni: Well, they can both be true, so the answer I think is _no_.  Think of them as `hasSomePrefixAttribute`, `hasSomeNonePrefixAttributes`. An element can have both types of attributes.

As far as I understand it, a prefix attribute is an id or class attribute that is inside the list of attributes and should move outside, to the element (e.g. `div(id="foo", class="bar")` becoming `#foo.bar`. And a non-prefix attribute is anything that stays inside the parentheses after formatting.

 I think the naming is actually fine as it is now? Or would you like to suggest better naming for these?

src/printer.ts
Shinigami92: I will let you decide ^^

src/printer.ts
lehni: You format on save, or you hate format on save?

src/printer.ts
lehni: Changed already! 😉

src/printer.ts
lehni: So this is the only remaining g open question here I think

lehni: I can add a comment that explains the above?

src/printer.ts
Shinigami92: I format on save / I love it (I even wrote a plugin to format for me on save 😏)

src/printer.ts
Shinigami92: Yes, this would be nice

lehni: I think I found a better way to do this now, see my latest commit. With this, the PR is finished from my side.

src/printer.ts
```diff
@@ -100,15 +100,11 @@ export class PugPrinter {
 	private result: string = '';
 
 	private currentIndex: number = 0;
+	private currentLineLength: number = 0;
```
lehni, line 103: @Shinigami92 I've moved this up here along with `currentIndex`, and I removed the JSDOC comment which wasn't actually meaningful for docs, and didn't say more then the code already does. And I've noticed one more thing: We are determining the length, not the actual position in a column, so this has to start at 0, meaning no content.

src/printer.ts
```diff
@@ -420,21 +416,24 @@ export class PugPrinter {
 			this.possibleClassPosition = this.result.length;
 			result = '(';
 			logger.debug(this.currentLineLength);
-			this.currentLineLength += 1;
```
lehni, line 417: @Shinigami92 I was able to merge the counting of the opening parenthesis together with the closing one further down. This made the code there simpler too, because I don't need to remove the already counted one if there aren't any attributes.

src/printer.ts
```diff
@@ -464,32 +464,29 @@ export class PugPrinter {
...
 				if (this.previousToken?.type === 'tag' && this.previousToken.val === 'div') {
 					this.currentLineLength -= 3;
 				}
 			}
-			if (hasNonPrefixAttributes) {
-				// Add trailing brace
-				this.currentLineLength += 1;
-			} else {
-				// Remove leading brace
-				this.currentLineLength -= 1;
```
lehni, line 470: @Shinigami92 here the part that got simpler, mentioned above.

lehni:
@Shinigami92 when looking into the handling of literal attributes, I found a new bug in the current implementation of the plugin that happens as soon as you have both an id and class attribute on an element. I already have a fix for this in a follow up PR, but it will require #121 to be merged first. And I don't want to include it here, to not complicate this PR further.

src/printer.ts
```diff
@@ -414,21 +416,24 @@ export class PugPrinter {
...
-			let nonPrefixAttributes: number = 0;
-			let hasPrefixAttribute: boolean = false;
-			let numAttributes: number = 0;
+			// In pug, tags can have two kind of attributes: normal attributes that appear between parantheses,
+			// and literals for ids and classes, prefixing the paranthesis, e.g.: `#id.class(attribute="value")`
+			// https://pugjs.org/language/attributes.html#class-literal
+			// https://pugjs.org/language/attributes.html#id-literal
+			// In the stream of attribute tokens, distinguish those that can be converted to literals,
+			// and count those that cannot (normal attributes) to determine the resulting line length correctly.
+			let hasLiteralAttributes: boolean = false;
```
Shinigami92, line 427: :heart:

Shinigami92 (approved):

PR #122, opened 2020-09-30
Update README to mention all pug alias options

lehni:
As discussed here: https://github.com/prettier/plugin-pug/issues/118#issuecomment-700724677

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,77 +77,34 @@ yarn add --dev prettier @prettier/plugin-pug
 yarn prettier --write "**/*.pug"
 ```
 
-### Prettier Options
+### Pug versions of standard prettier options
 
-- `bracketSpacing` | `pugBracketSpacing`  
-  If you want to configure different bracketSpacing for pug than for js code, you can use prettier's override.
-  ```json
-  {
-    "bracketSpacing": true,
-    "overrides": [
-      {
-        "files": "*.pug",
-        "options": {
-          "parser": "pug",
-          "bracketSpacing": false
-        }
-      }
-    ]
-  }
-  ```
-- `printWidth` | `pugPrintWidth`  
-  _Currently not very accurate, but works_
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -200,7 +202,8 @@ They should be set via `Prettier`'s `overrides` option
   - `'trim-all'` -> Trim all spaces within comments.  
     Example: `// this is a comment`
 
-- `pugSortAttributesBeginning` (and `pugSortAttributesEnd`)
+- `sortAttributesBeginning` | `pugSortAttributesBeginning`
```
Shinigami92, line 205: There is (intentionally) no `sortAttributesBeginning` or `sortAttributesEnd`
I also think about to deprecate all previous options without `pug` prefix that are not a prettier default option

Shinigami92 (changes requested):

README.md
lehni: Ah right, you mentioned that! Will remove it. Shall I remove the other also, so people stop using them?

Shinigami92: I think yes, we should remove the non alias ones (pug-specific-option)

lehni: Ok, change coming up!

lehni: But then the whole part about __They should be set via `Prettier`'s `overrides` option__ can be left out, right? Making docs simpler, a good thing!

Shinigami92: Yes, this also can be upgraded to just use the new pug prefixes

You only need to mention specifically that `singleQuote` will be the opposite of what prettier attributes will be quoted
Cause the js code in values of pug attributes uses these quotes

lehni: Yes makes sense!

Shinigami92 (approved):

lehni:
Thanks for the merge, and apologies for the spelling mistakes that made it through 🙈

PR #123, opened 2020-09-30
Add docs for pugWrapAttributesThreshold to README

lehni:


README.md
```diff
@@ -152,18 +152,52 @@ These additional options are specific to pug templates and can be configured in
...
+    input(type="text")
+    input(
+      type="text",
+      value="my_value",
+      name="my_name"
+    )
+    ```
+  - `2 … Infinity` → Same as above, just with different thresholds.  
 
 The definitions for these options can be found in [src/options/index.ts](https://github.com/prettier/plugin-pug/blob/master/src/options/index.ts)
```
Shinigami92, line 202: Could you also change this? `src/options/index.ts` is outdated :thinking: 
Maybe `src/options/*`? Or leave it completely?!

README.md
```diff
@@ -109,11 +109,11 @@ These additional options are specific to pug templates and can be configured in
 
   Choices:
 
-  - `'always'` _default_ -> Always separate attributes with commas.  
+  - `'always'` _default_ → Always separate attributes with commas.
```
Shinigami92, line 112: I'm a little bit afraid of these arrows ^^'
Do thy work on any OS and always will be rendered correctly?

README.md
```diff
@@ -152,18 +152,52 @@ These additional options are specific to pug templates and can be configured in
...
+    Example:
+    ```pug
+    input(type="text")
+    input(
+      type="text",
+      value="my_value",
+      name="my_name"
+    )
+    ```
+  - `2 … Infinity` → Same as above, just with different thresholds.
```
Shinigami92, line 200: Same for this `…`
Also IMO ranges are written with only two dots 🤣 `2 .. Infinity`

Shinigami92 (changes requested):

README.md
lehni: Good point. I think they're fine in 2020, but happy to switch back :)

README.md
lehni: right! I'll switch

README.md
Shinigami92: ok, leave it out

README.md
lehni: But that's math notation then, and would need `[2 .. Infinity]`, looking like an array ;)

Shinigami92: 😏

README.md
lehni: Done!

Shinigami92 (approved):

PR #125, opened 2020-10-02
Support prettier-ignore comments inside pug templates

lehni:
Use `//- prettier-ignore` comments to prevent automatic formatting of the following line or block:
```pug
div
  .wrapper-1(attributes1="foo", :attribute2="bar", attribute3="something too long to keep on one line")
  //- prettier-ignore
  .wrapper-2(attributes1="foo", :attribute2="bar", attribute3="something too long to keep on one line")
  .wrapper-3(attributes1="foo", :attribute2="bar", attribute3="something too long to keep on one line")
  //- prettier-ignore
  div
    .wrapper-4(attributes1="foo", :attribute2="bar", attribute3="something too long to keep on one line")
.wrapper-5(attributes1="foo", :attribute2="bar", attribute3="something too long to keep on one line")
```
->
```pug
div
  .wrapper-1(
    attributes1="foo",
    :attribute2="bar",
    attribute3="something too long to keep on one line"
  )
  //- prettier-ignore
  .wrapper-2(attributes1="foo", :attribute2="bar", attribute3="something too long to keep on one line")
  .wrapper-3(
    attributes1="foo",
    :attribute2="bar",
    attribute3="something too long to keep on one line"
  )
  //- prettier-ignore
  div
    .wrapper-4(attributes1="foo", :attribute2="bar", attribute3="something too long to keep on one line")
.wrapper-5(
  attributes1="foo",
  :attribute2="bar",
  attribute3="something too long to keep on one line"
)
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,3 +13,7 @@ max_line_length = off
 trim_trailing_whitespace = false
 indent_style = space
 indent_size = 2
+
+[*.pug]
+indent_style = space
+indent_size = 2
--- a/README.md
+++ b/README.md
@@ -33,7 +33,8 @@ Please note that the [plugin ecosystem in Prettier](https://prettier.io/docs/en/
 
 Plugin for Prettier to format pug code
 
-You can disable code formatting for a particular code block by adding <nobr>`<!-- prettier-ignore -->`</nobr> before ` ```pug `.
+You can disable code formatting for a particular code block by adding <nobr>`<!-- prettier-ignore -->`</nobr> before ` ```pug `,
+as well as by using <nobr>`//- prettier-ignore`</nobr> comments inside your pug templates.
 
 ````markdown
 Pug code with custom formatting:
--- a/README.md
+++ b/README.md
@@ -66,9 +66,9 @@ You can disable code formatting for a particular element by adding <nobr>`//- pr
 
 ```pug
 div.text( color =   "primary",  disabled  ="true"  )
-// -prettier-ignore
+//- prettier-ignore
 div.text( color =   "primary",  disabled  ="true"  )
-// -prettier-ignore
+//- prettier-ignore: because of reasons
 div
   div.text( color =   "primary",  disabled  ="true"  )
 ```
@@ -77,9 +77,9 @@ Prettified output:
 
 ```pug
 .text(color="primary", disabled)
-// -prettier-ignore
+//- prettier-ignore
 div.text( color =   "primary",  disabled  ="true"  )
-// -prettier-ignore
+//- prettier-ignore: because of reasons
 div
   div.text( color =   "primary",  disabled  ="true"  )
 ```
--- a/README.md
+++ b/README.md
@@ -33,24 +33,6 @@ Please note that the [plugin ecosystem in Prettier](https://prettier.io/docs/en/
 
 Plugin for Prettier to format pug code
 
-You can disable code formatting for a particular code block by adding <nobr>`<!-- prettier-ignore -->`</nobr> before ` ```pug `,
-as well as by using <nobr>`//- prettier-ignore`</nobr> comments inside your pug templates.
-
-````markdown
-Pug code with custom formatting:
-
-<!-- prettier-ignore -->
-```pug
-div.text( color =   "primary",  disabled  ="true"  )
-```
-
-Prettified code:
-
-```pug
-.text(color="primary", disabled)
-```
-````
-
 ## Getting started
 
 Simply install `prettier` and `@prettier/plugin-pug` as your project’s npm devDependencies:
@@ -77,6 +59,47 @@ yarn add --dev prettier @prettier/plugin-pug
 ## or
 yarn prettier --write "**/*.pug"
 ```
+
+### Selectively ignoring automatic formatting
+
+You can disable code formatting for a particular element by adding <nobr>`//- prettier-ignore`</nobr> comments in your pug templates:
+
+```pug
+div.text( color =   "primary",  disabled  ="true"  )
+// -prettier-ignore
+div.text( color =   "primary",  disabled  ="true"  )
+// -prettier-ignore
+div
+  div.text( color =   "primary",  disabled  ="true"  )
+```
+
+Prettified output:
+
+```pug
+.text(color="primary", disabled)
+// -prettier-ignore
+div.text( color =   "primary",  disabled  ="true"  )
+// -prettier-ignore
+div
+  div.text( color =   "primary",  disabled  ="true"  )
+```
+
+You can also disable code formatting in Markdown for a particular ` ```pug ` block by adding <nobr>`<!-- prettier-ignore -->`</nobr> before the block:
+
+````markdown
+Pug code with preserved custom formatting:
+
+<!-- prettier-ignore -->
+```pug
+div.text( color =   "primary",  disabled  ="true"  )
+```
+
+Pug code with automatic formatting:
+
+```pug
+.text(color="primary", disabled)
+```
+````
 
 ### Pug versions of standard prettier options
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -23,8 +23,8 @@ if (process.env.NODE_ENV === 'test') {
 }
 
 type FastPathStackEntry = {
-	content: string,
-	tokens: Token[]
+	content: string;
+	tokens: Token[];
 };
 
 export const plugin: Plugin = {
--- a/src/index.ts
+++ b/src/index.ts
@@ -22,6 +22,11 @@ if (process.env.NODE_ENV === 'test') {
 	logger.setLogLevel(LogLevel.DEBUG);
 }
 
+type FastPathStackEntry = {
+	content: string,
+	tokens: Token[]
+};
+
 export const plugin: Plugin = {
 	languages: [
 		{
@@ -38,13 +43,14 @@ export const plugin: Plugin = {
 	],
 	parsers: {
 		pug: {
-			parse(text: string, parsers: { [parserName: string]: Parser }, options: ParserOptions): Token[] {
+			parse(text: string, parsers: { [parserName: string]: Parser }, options: ParserOptions): FastPathStackEntry {
 				logger.debug('[parsers:pug:parse]:', { text });
-				const tokens: lex.Token[] = lex(text.trimLeft());
+				const content: string = text.trimLeft();
+				const tokens: lex.Token[] = lex(content);
 				// logger.debug('[parsers:pug:parse]: tokens', JSON.stringify(tokens, undefined, 2));
 				// const ast: AST = parse(tokens, {});
 				// logger.debug('[parsers:pug:parse]: ast', JSON.stringify(ast, undefined, 2));
-				return tokens;
+				return { content, tokens };
 			},
 			astFormat: 'pug-ast',
 			hasPragma(text: string): boolean {
@@ -67,9 +73,10 @@ export const plugin: Plugin = {
 	printers: {
 		'pug-ast': {
 			print(path: FastPath, options: ParserOptions & PugParserOptions, print: (path: FastPath) => Doc): Doc {
-				const tokens: Token[] = path.stack[0];
+				const entry: FastPathStackEntry = path.stack[0];
+				const { content, tokens } = entry;
 				const pugOptions: PugPrinterOptions = convergeOptions(options);
-				const printer: PugPrinter = new PugPrinter(tokens, pugOptions);
+				const printer: PugPrinter = new PugPrinter(content, tokens, pugOptions);
 				const result: string = printer.build();
 				logger.debug('[printers:pug-ast:print]:', result);
 				return result;
--- a/src/printer.ts
+++ b/src/printer.ts
@@ -771,7 +771,7 @@ export class PugPrinter {
 
 	private comment(commentToken: CommentToken): string {
 		let result: string = this.computedIndent;
-		if (commentToken.val.trim() === 'prettier-ignore') {
+		if (/^\s*prettier-ignore\b/.test(commentToken.val)) {
 			// Use a own token processing loop to find the end of the stream of tokens to be ignored by formatting,
 			// and uses their `loc` properties to retrieve the original pug code to be used instead.
 			let token: Token | null = this.getNextToken();
--- a/src/printer.ts
+++ b/src/printer.ts
@@ -99,9 +99,10 @@ export interface PugPrinterOptions {
 export class PugPrinter {
 	private result: string = '';
 
-	private currentIndex: number = 0;
+	// The current index in the `tokens` array, starts at -1, because `getNextToken()` increases before retrieving
+	private currentIndex: number = -1;
 	private currentLineLength: number = 0;
-
+	
 	private readonly indentString: string;
 	private indentLevel: number = 0;
 
@@ -152,14 +153,6 @@ export class PugPrinter {
 		};
 	}
 
-	private get previousToken(): Token | undefined {
-		return this.tokens[this.currentIndex - 1];
-	}
-
-	private get nextToken(): Token | undefined {
-		return this.tokens[this.currentIndex + 1];
-	}
-
 	public build(): string {
 		const results: string[] = [];
 		if (this.tokens[0]?.type === 'text') {
@@ -167,10 +160,8 @@ export class PugPrinter {
 		} else if (this.tokens[0]?.type === 'eos') {
 			return '';
 		}
-		let index: number = 0;
-		while (index < this.tokens.length) {
-			this.currentIndex = index;
-			let token: Token = this.tokens[index++];
+		let token: Token | null = this.getNextToken();
+		while (token) {
 			logger.debug('[PugPrinter]:', JSON.stringify(token));
 			try {
 				switch (token.type) {
@@ -203,47 +194,53 @@ export class PugPrinter {
 					token.type === 'comment' &&
 					token.val.trim() === 'prettier-ignore'
 				) {
-					// Use a separate token processing loop that finds the end of the tokens to be ignored by formatting,
+					// Use a own token processing loop to find the end of the stream of tokens to be ignored by formatting,
 					// and uses their `loc` properties to retrieve the original pug code to be used instead.
-					const firstToken: Token = this.tokens[index];
-					const startsWithNewLine: boolean = firstToken.type === 'newline';
-					let skipNewline: boolean = startsWithNewLine;
-					let ignoreLevel: number = 0;
-					while (index < this.tokens.length) {
-						token = this.tokens[index++];
-						const { type } = token;
-						if (type === 'newline' && ignoreLevel === 0) {
-							// Skip first newline after prettier-ignore comment:
-							if (skipNewline) {
-								skipNewline = false;
-							} else {
-								break;
+					token = this.getNextToken();
+					if (token) {
+						const firstToken: Token = token;
+						const startsWithNewLine: boolean = firstToken.type === 'newline';
+						let skipNewline: boolean = startsWithNewLine;
+						let ignoreLevel: number = 0;
+						while (token) {
+							const { type } = token;
+							if (type === 'newline' && ignoreLevel === 0) {
+								// Skip first newline after prettier-ignore comment:
+								if (skipNewline) {
+									skipNewline = false;
+								} else {
+									break;
+								}
+							} else if (type === 'indent') {
+								ignoreLevel++;
+							} else if (type === 'outdent') {
+								ignoreLevel--;
+								if (ignoreLevel === 0) {
+									break;
+								}
 							}
-						} else if (type === 'indent') {
-							ignoreLevel++;
-						} else if (type === 'outdent') {
-							ignoreLevel--;
-							if (ignoreLevel === 0) {
-								break;
+							token = this.getNextToken();
+						}
+						if (token) {
+							const lastToken: Token = token;
+							const lines: string[] = this.getUnformattedContentLines(firstToken, lastToken);
+							// Start with an empty string for the first newline after comment.
+							if (startsWithNewLine) {
+								lines.unshift('');
 							}
+							// Trim the last line, since indentation of formatted pug is handled separately.
+							const lastLine: string | undefined = lines.pop();
+							if (lastLine !== undefined) {
+								lines.push(lastLine.trimRight());
+							}
+							results.push(lines.join('\n'));
 						}
 					}
-					const lastToken: Token = this.tokens[index - 1];
-					const lines: string[] = this.getUnformattedContentLines(firstToken, lastToken);
-					// Start with an empty string for the first newline after comment.
-					if (startsWithNewLine) {
-						lines.unshift('');
-					}
-					// Trim the last line, since indentation of formatted pug is handled separately.
-					const lastLine: string | undefined = lines.pop();
-					if (lastLine !== undefined) {
-						lines.push(lastLine.trimRight());
-					}
-					results.push(lines.join('\n'));
 				}
 			} catch {
 				throw new Error('Unhandled token: ' + JSON.stringify(token));
 			}
+			token = this.getNextToken();
 		}
 		return results.join('');
 	}
@@ -267,6 +264,19 @@ export class PugPrinter {
 				return this.indentString;
 		}
 		return '';
+	}
+
+	private get previousToken(): Token | undefined {
+		return this.tokens[this.currentIndex - 1];
+	}
+
+	private get nextToken(): Token | undefined {
+		return this.tokens[this.currentIndex + 1];
+	}
+
+	private getNextToken(): Token | null {
+		this.currentIndex++;
+		return this.tokens[this.currentIndex] ?? null;
 	}
 
 	private quoteString(val: string): string {
--- a/src/printer.ts
+++ b/src/printer.ts
@@ -129,7 +129,7 @@ export class PugPrinter {
 	private pipelessComment: boolean = false;
 
 	public constructor(
-		private content: string,
+		private readonly content: string,
 		private tokens: Token[],
 		private readonly options: PugPrinterOptions
 	) {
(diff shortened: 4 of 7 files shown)


lehni:
@Shinigami92 I tried to request a review but for some reason, the button doesn't show up on this PR

lehni:
It's up for debate if it only ever should ignore the next line, or all nested elements as well (e.g. the 2nd example above, the `div` with nested `.wrapper-4`). Happy to discuss!

Shinigami92:
> @Shinigami92 I tried to request a review but for some reason, the button doesn't show up on this PR

Ok I need to investigate why my community cannot request these. Maybe it's a GitHub Teams feature to do so 🤔 
Or I need to relax permissions somewhere

lehni:
@Shinigami92 this used to work, I think? Maybe it was something you changed recently?

Shinigami92:
> It's up for debate if it only ever should ignore the next line, or all nested elements as well (e.g. the 2nd example above, the `div` with nested `.wrapper-4`).
> 
> Personally I find it more pug-like if it affects the nested things as well. That's also how commenting things out works in pug.
> 
> Happy to discuss!

Very interesting idea 🤔 
But is it hard to implement?

_I mean I need to look in your current working code_ 😅

lehni:
The current implementation does that already!

README.md
```diff
@@ -33,7 +33,8 @@ Please note that the [plugin ecosystem in Prettier](https://prettier.io/docs/en/
 
 Plugin for Prettier to format pug code
 
-You can disable code formatting for a particular code block by adding <nobr>`<!-- prettier-ignore -->`</nobr> before ` ```pug `.
+You can disable code formatting for a particular code block by adding <nobr>`<!-- prettier-ignore -->`</nobr> before ` ```pug `,
+as well as by using <nobr>`//- prettier-ignore`</nobr> comments inside your pug templates.
```
Shinigami92, line 37: Could you swap these sentence parts with each other? So `//- prettier-ignore` becomes more priority

Shinigami92 (changes requested):

README.md
lehni: Yes good point. BTW, the example underneath is wrong, no?

```
<!-- prettier-ignore -->
\`\`\`pug
div.text( color =   "primary",  disabled  ="true"  )
\`\`\`

Prettified code:

\`\`\`pug
.text(color="primary", disabled)
```
\`\`\`

lehni: Using underscores because I don't know how to nest code in code ;)

lehni: The point is: The Prettified code shouldn't actually be prettified in the example?

Shinigami92:
Fucking hell, can I come across the globe and just hug you?!
Damn it, this is rocket science

I really start to think about to give both of you (also @SkyaTura) more privileges in my repo

But, need to lock into it with IDE in depth
Will go home now and work further from there 😃

README.md
Shinigami92: No, it's all correct
The `<!-- prettier-ignore -->` only applies to its own "```" block

README.md
```diff
@@ -33,20 +33,30 @@ Please note that the [plugin ecosystem in Prettier](https://prettier.io/docs/en/
...
 Plugin for Prettier to format pug code
 
-You can disable code formatting for a particular code block by adding <nobr>`<!-- prettier-ignore -->`</nobr> before ` ```pug `.
+You can disable code formatting for a particular code block by adding <nobr>`//- prettier-ignore`</nobr> comments inside your pug templates
+or by adding <nobr>`<!-- prettier-ignore -->`</nobr> before ` ```pug `.
 
 ````markdown
 Pug code with custom formatting:
 
-<!-- prettier-ignore -->
```
Shinigami92, line 41: This was part of the example ^^'
The example for how to use it in `md`

Shinigami92 (changes requested):

README.md
lehni: Ouff, sorry! I don't understand what it's evem doing then?

README.md
lehni: Let's keep the discussion in the thread above with the rest of the context?

README.md
Shinigami92: I want to present an example of how it would look like to use non-formatting in markdown files

Using it in pug itself should be its own example block :slightly_smiling_face:

Shinigami92: That's also the reason for the four "\`" instead of three "`"

lehni: Is this about using it with markdown?

Shinigami92: yes

README.md
lehni: I've pushed a new version of the README with its own chapter for Selectively ignoring automatic formatting

Shinigami92:
@lehni and @SkyaTura: If you didn't already noticed ->
I gave both of you `Role: Write` privileges to my repo :tada:  
So now you can delete your forks (when I have merged all current PRs from you), and checkout this origin directly
You both cannot work on `master`-branch but you can freely create your own branches in this repo

Also I think, you can add and change some labels and assignees etc in issues and PRs :thinking:

lehni:
I'm away from the computer now. Will write more later. Thanks!

src/index.ts
```diff
@@ -22,6 +22,11 @@ if (process.env.NODE_ENV === 'test') {
 	logger.setLogLevel(LogLevel.DEBUG);
 }
 
+type FastPathStackEntry = {
+	content: string,
```
Shinigami92, line 26:
```suggestion
	content: string;
```

src/index.ts
```diff
@@ -22,6 +22,11 @@ if (process.env.NODE_ENV === 'test') {
...
+	tokens: Token[]
```
Shinigami92, line 27:
```suggestion
	tokens: Token[];
```

src/printer.ts
```diff
@@ -194,6 +199,48 @@ export class PugPrinter {
 						results.push(this[token.type](token));
 						break;
 				}
+				if (
+					token.type === 'comment' &&
+					token.val.trim() === 'prettier-ignore'
+				) {
+					// Use a spearate token processing loop that finds the end of the tokens to be ignored by formatting,
```
Shinigami92, line 206:
```suggestion
					// Use a separate token processing loop that finds the end of the tokens to be ignored by formatting,
```

src/printer.ts
```diff
@@ -194,6 +199,48 @@ export class PugPrinter {
...
+					// and uses their `loc` properties to retreive the original pug code to be used instead.
```
Shinigami92, line 207:
```suggestion
					// and uses their `loc` properties to retrieve the original pug code to be used instead.
```

src/printer.ts
```diff
@@ -194,6 +199,48 @@ export class PugPrinter {
...
+					const firstToken = this.tokens[index];
```
Shinigami92, line 208:
```suggestion
					const firstToken: Token = this.tokens[index];
```

src/printer.ts
```diff
@@ -194,6 +199,48 @@ export class PugPrinter {
...
+					const startsWithNewLine = firstToken.type === 'newline';
```
Shinigami92, line 209:
```suggestion
					const startsWithNewLine: boolean = firstToken.type === 'newline';
```

src/printer.ts
```diff
@@ -194,6 +199,48 @@ export class PugPrinter {
...
+					let skipNewline = startsWithNewLine;
```
Shinigami92, line 210:
```suggestion
					let skipNewline: boolean = startsWithNewLine;
```

src/printer.ts
```diff
@@ -194,6 +199,48 @@ export class PugPrinter {
...
+					let ignoreLevel = 0;
```
Shinigami92, line 211:
```suggestion
					let ignoreLevel: number = 0;
```

src/printer.ts
```diff
@@ -194,6 +199,48 @@ export class PugPrinter {
...
+						} else if (type === 'indent') {
+							ignoreLevel++;
+						} else if (type === 'outdent') {
+							ignoreLevel--;
+							if (ignoreLevel === 0) {
+								break;
+							}
+						}
+					}
+					const lastToken = this.tokens[index - 1];
```
Shinigami92, line 231:
```suggestion
					const lastToken: Token = this.tokens[index - 1];
```

src/printer.ts
```diff
@@ -194,6 +199,48 @@ export class PugPrinter {
...
+					const lines = this.getUnformattedContentLines(firstToken, lastToken);
```
Shinigami92, line 232:
```suggestion
					const lines: string[] = this.getUnformattedContentLines(firstToken, lastToken);
```

src/printer.ts
```diff
@@ -194,6 +199,48 @@ export class PugPrinter {
...
+					// Start with an empty string for the first newline after comment.
+					if (startsWithNewLine) {
+						lines.unshift('');
+					}
+					// Trim the last line, since indentation of formatted pug is handled separately.
+					const lastLine = lines.pop();
```
Shinigami92, line 238:
```suggestion
					const lastLine: string | undefined = lines.pop();
```

src/printer.ts
```diff
@@ -230,6 +277,21 @@ export class PugPrinter {
 		return !!token && possibilities.includes(token.type) !== invert;
 	}
 
+	private getUnformattedContentLines(firstToken: Token, lastToken: Token): string[] {
+		const { start } = firstToken.loc;
+		const { end } = lastToken.loc;
+		const lines = this.content.split(/\r\n|\n|\r/);
```
Shinigami92, line 283:
```suggestion
		const lines: string[] = this.content.split(/\r\n|\n|\r/);
```

src/printer.ts
```diff
@@ -230,6 +277,21 @@ export class PugPrinter {
...
+		const startLine = start.line - 1;
```
Shinigami92, line 284:
```suggestion
		const startLine: number = start.line - 1;
```

src/printer.ts
```diff
@@ -230,6 +277,21 @@ export class PugPrinter {
...
+		const endLine = end.line - 1;
```
Shinigami92, line 285:
```suggestion
		const endLine: number = end.line - 1;
```

src/printer.ts
```diff
@@ -230,6 +277,21 @@ export class PugPrinter {
...
+		const parts: string[] = [];
+		parts.push(lines[startLine].substring(start.column - 1));
+		for (let line = startLine + 1; line < endLine; line++) {
```
Shinigami92, line 288:
```suggestion
		for (let line: number = startLine + 1; line < endLine; line++) {
```

Shinigami92 (changes requested):

src/printer.ts
```diff
@@ -163,9 +167,10 @@ export class PugPrinter {
 		} else if (this.tokens[0]?.type === 'eos') {
 			return '';
 		}
-		for (let index: number = 0; index < this.tokens.length; index++) {
+		let index: number = 0;
+		while (index < this.tokens.length) {
 			this.currentIndex = index;
-			const token: Token = this.tokens[index];
+			let token: Token = this.tokens[index++];
```
Shinigami92, line 173: Please move this `index++` out of it's usage, so a reader can directly see if first or last the increment was made

So it will be:
```ts
index++;
let token: Token = this.tokens[index];
```
or
```ts
let token: Token = this.tokens[index];
index++;
```

src/printer.ts
```diff
@@ -194,6 +199,48 @@ export class PugPrinter {
 						results.push(this[token.type](token));
 						break;
 				}
+				if (
+					token.type === 'comment' &&
```
Shinigami92, line 203: Can all this logic maybe moved to `comment` member-function down in this file? Or is there a reason why this should be done here? Maybe cause you accessing the `index`? :thinking:  

https://github.com/prettier/plugin-pug/blob/48dfff17416f21bc3fe661d2e4a8d9265604e2f9/src/printer.ts#L736

src/printer.ts
```diff
@@ -194,6 +199,48 @@ export class PugPrinter {
...
+					// and uses their `loc` properties to retrieve the original pug code to be used instead.
+					const firstToken: Token = this.tokens[index];
+					const startsWithNewLine: boolean = firstToken.type === 'newline';
+					let skipNewline: boolean = startsWithNewLine;
+					let ignoreLevel: number = 0;
+					while (index < this.tokens.length) {
+						token = this.tokens[index++];
+						const { type } = token;
+						if (type === 'newline' && ignoreLevel === 0) {
+							// Skip first newline after prettier-ignore comment:
```
Shinigami92, line 216:
```suggestion
							// Skip first newline after `prettier-ignore` comment
```

src/printer.ts
```diff
@@ -194,6 +199,48 @@ export class PugPrinter {
...
+					token.val.trim() === 'prettier-ignore'
+				) {
+					// Use a separate token processing loop that finds the end of the tokens to be ignored by formatting,
+					// and uses their `loc` properties to retrieve the original pug code to be used instead.
+					const firstToken: Token = this.tokens[index];
+					const startsWithNewLine: boolean = firstToken.type === 'newline';
+					let skipNewline: boolean = startsWithNewLine;
+					let ignoreLevel: number = 0;
+					while (index < this.tokens.length) {
+						token = this.tokens[index++];
```
Shinigami92, line 213: Same as above

src/printer.ts
```diff
@@ -194,6 +199,48 @@ export class PugPrinter {
...
+					token.type === 'comment' &&
+					token.val.trim() === 'prettier-ignore'
+				) {
+					// Use a separate token processing loop that finds the end of the tokens to be ignored by formatting,
+					// and uses their `loc` properties to retrieve the original pug code to be used instead.
+					const firstToken: Token = this.tokens[index];
+					const startsWithNewLine: boolean = firstToken.type === 'newline';
+					let skipNewline: boolean = startsWithNewLine;
+					let ignoreLevel: number = 0;
+					while (index < this.tokens.length) {
```
Shinigami92, line 212: This makes me curious :thinking:  
You are accessing and also **mutate** the same `index` as in the outer while loop?

Does this always work? Or did you just catch special cases with your current tests?

Shinigami92 (changes requested):

Shinigami92:
So if I understood everything correct, the following wouldn't make sense, and therefore is correctly commented?

```pug
div
  //- prettier-ignore | (1)
  div
    //- prettier-ignore | <-- already ignored be the above one (1)?!
    div
  div
    div
    //- prettier-ignore | (2)
    div
      //- prettier-ignore | <-- already ignored be the above one (2)?!
      div
  div
```

---

Should we allow a _reason_?

```pug
div
  //- prettier-ignore: I want this this way!
  div
```

lehni:
> So if I understood everything correct, the following wouldn't make sense, and therefore is correctly commented?
> 
> ```pug
> div
>   //- prettier-ignore | (1)
>   div
>     //- prettier-ignore | <-- already ignored be the above one (1)?!
>     div
>   div
>     div
>     //- prettier-ignore | (2)
>     div
>       //- prettier-ignore | <-- already ignored be the above one (2)?!
>       div
>   div
> ```

Yes, exactly! It's kind of similar to what happens when you comment out a block of pug code:

```pug
// everything indented underneath will be commented out:
  div
    div
  div
    div
    div
      div
  div
```
 
> Should we allow a _reason_?
> 
> ```pug
> div
>   //- prettier-ignore: I want this this way!
>   div
> ```

Yes, we could simply match `/^\s*prettier-ignore\b/ and whatever follows is up to you.

src/printer.ts
lehni: I was trying this approach first, but yes the problem is that it has to increase the index and skip all the processing of all the other tokens, so it seemed easer to do it outside of the normal processing loop.

src/printer.ts
lehni: Yes this works, and it's the reason why I switched to `while` loops from `for`. The looks themselves just check the condition, they don't increase the index. You can have an inner loop that increases the same index as the outer loop. The inner loop usually stops because of one of the `break` statements inside of it. The `while` condition is just there to make sure it never loops infinitely.  You could replace it with `while(true) {` and it would still work.

src/printer.ts
lehni: ok

src/printer.ts
lehni: I will try to add a `getNextToken()` helper that internally increases `this.currentIndex` so that the whole thing becomes a bit more like reading from a stream. And the loop and nested loop can then both iterate until this helper returns `null`.

Does this sound better?

src/printer.ts
lehni: I found a way to make this work! Check the last two commits, I think it's very clean now.

src/printer.ts
lehni: I've managed to implement it exactly like this, and it works well I think!

.editorconfig
```diff
@@ -13,3 +13,7 @@ max_line_length = off
 trim_trailing_whitespace = false
 indent_style = space
 indent_size = 2
+
+[*.pug]
```
Shinigami92, line 17: Not drastically, but could you explain why you added this?

README.md
```diff
@@ -77,6 +60,47 @@ yarn add --dev prettier @prettier/plugin-pug
 yarn prettier --write "**/*.pug"
 ```
 
+### Selectively ignoring automatic formatting
+
+You can disable code formatting for a particular element by adding <nobr>`//- prettier-ignore`</nobr> comments in your pug templates:
+
+```pug
+div.text( color =   "primary",  disabled  ="true"  )
+// -prettier-ignore
```
Shinigami92, line 69:
```suggestion
//- prettier-ignore
```

README.md
```diff
@@ -77,6 +60,47 @@ yarn add --dev prettier @prettier/plugin-pug
...
+div.text( color =   "primary",  disabled  ="true"  )
+// -prettier-ignore
```
Shinigami92, line 71:
```suggestion
//- prettier-ignore: just leave it as is
```

If you have a better _reason_ sentence, you are free to use it :slightly_smiling_face:

README.md
```diff
@@ -77,6 +60,47 @@ yarn add --dev prettier @prettier/plugin-pug
...
+div
+  div.text( color =   "primary",  disabled  ="true"  )
+```
+
+Prettified output:
+
+```pug
+.text(color="primary", disabled)
+// -prettier-ignore
```
Shinigami92, line 80:
```suggestion
//- prettier-ignore
```

README.md
```diff
@@ -77,6 +60,47 @@ yarn add --dev prettier @prettier/plugin-pug
...
+div.text( color =   "primary",  disabled  ="true"  )
+// -prettier-ignore
```
Shinigami92, line 82:
```suggestion
//- prettier-ignore: just leave it as is
```

Shinigami92 (changes requested):

src/printer.ts
```diff
@@ -98,7 +98,8 @@ export interface PugPrinterOptions {
 export class PugPrinter {
 	private result: string = '';
 
-	private currentIndex: number = 0;
+	// The current index in the `tokens` array, starts at -1, because `getNextToken()` increases before retrieving
```
Shinigami92, line 101: Could you convert this to JSDoc?
So on hover we can see the comment.
Don't know if we should split the descriptive part from the developer comment part :thinking:

src/printer.ts
```diff
@@ -230,6 +240,21 @@ export class PugPrinter {
...
 	}
 
+	private getUnformattedContentLines(firstToken: Token, lastToken: Token): string[] {
+		const { start } = firstToken.loc;
+		const { end } = lastToken.loc;
+		const lines: string[] = this.content.split(/\r\n|\n|\r/);
+		const startLine: number = start.line - 1;
+		const endLine: number = end.line - 1;
+		const parts: string[] = [];
+		parts.push(lines[startLine].substring(start.column - 1));
```
Shinigami92, line 250:
```suggestion
		parts.push(lines[startLine].slice(start.column - 1));
```

src/printer.ts
```diff
@@ -230,6 +240,21 @@ export class PugPrinter {
...
+		for (let line: number = startLine + 1; line < endLine; line++) {
+			parts.push(lines[line]);
+		}
+		parts.push(lines[endLine].substring(0, end.column - 1));
```
Shinigami92, line 254:
```suggestion
		parts.push(lines[endLine].slice(0, end.column - 1));
```

Shinigami92 (changes requested):

src/printer.ts
```diff
@@ -733,18 +758,56 @@ export class PugPrinter {
...
-			result += ' ';
-		}
-		result += '//';
-		if (!token.buffer) {
-			result += '-';
-		}
-		result += formatCommentPreserveSpaces(token.val, this.options.commentPreserveSpaces);
-		if (this.nextToken?.type === 'start-pipeless-text') {
-			this.pipelessComment = true;
+		if (/^\s*prettier-ignore\b/.test(commentToken.val)) {
```
Shinigami92, line 763: We should try to stick to the same behaviour as [typescript-eslint does with ban-ts-comment allow-with-description](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-ts-comment.md#allow-with-description)

tests/prettier-ignore/prettier-ignore.test.ts
```diff
@@ -0,0 +1,14 @@
+import { readFileSync } from 'fs';
+import { resolve } from 'path';
+import { format } from 'prettier';
+import { plugin } from './../../src/index';
+
+describe('prettier-ignore', () => {
+	test('should handle // prettier-ignore statements', () => {
```
Shinigami92, line 7:
```suggestion
	test("should handle '//- prettier-ignore' statements", () => {
```

Shinigami92 (changes requested):

Shinigami92:
I really like that feature, but we should check this against a code base with multiple `//- prettier-ignore` on some places and watch the performance  
Can you test this somewhere?

.editorconfig
lehni: I am using VSCode with the editorconfig plugin, and whenever I edited pug files, it used 4 spaces indentation with tabs instead of spaces. It even changed existing formatting on save.

README.md
lehni: Oops!

src/printer.ts
lehni: `line` is a string, not an array?

src/printer.ts
lehni: ah good, wasn't aware of this. Will check!

src/printer.ts
Shinigami92: YES! I know: https://stackoverflow.com/questions/2243824/what-is-the-difference-between-string-slice-and-string-substring

lehni:
@Shinigami92 I am happy to add more tests for various scenarios of nesting, was thinking the same!

But I don't think we need a performance test, and that would be hard to even test reliably, because:

When the `build()`loop encounters a `prettier-ignore` comment, it then uses a simple loop in `comment()` to find the end of the stream of tokens to skip / leave unformatted. That loop uses a very simple logic that is faster than the normal processing of these tokens would be. The total amount of tokens treated in one loop or the other doesn't change. Once the code in `comment()` has found the range of tokens to leave unformatted, it uses the `loc` property of the first and the last token to simply get the unformatted string from the original pug content. This again isn't expensive, no complexity involved. (see: `getUnformattedContentLines()`). 

Bottom line: pug code that contains `prettier-ignore` comments will most likely process faster than the same code without those comments, because there is less formatting and processing involved, not more.

lehni:
I'm away from the computer mostly today, will wrap this up tomorrow probably! Thanks for the reviews @Shinigami92!

Shinigami92:
@lehni I didn't mean an automated test, I just meant a one time _study_
So you use `yarn link` to link this repo against a code base using the feature and then see if the processing of `//- prettier-ignore` takes much longer as not processing it
And if so, we should write a warning note into readme, to not use this feature heavily

src/printer.ts
lehni: Crazy! All these years of writing JS and I never noticed that strings have a `slice()` method that does exactly the same as `substring()`. But why adding both? So strange 🤣

README.md
lehni: I quite like `  //- prettier-ignore: because reasons` :wink

lehni: https://mltshp-cdn.com/r/EXAW

lehni: EXAW

lehni:
@Shinigami92 I did a little speed test now, using my improved test in `prettier-ignore.test.js`, with some simple modifications to run `format()` 100 times instead of once, and measure the time:

```js
const expected: string = readFileSync(resolve(__dirname, 'formatted.pug'), 'utf8');
const code: string = readFileSync(resolve(__dirname, 'unformatted.pug'), 'utf8');
let actual: string = '';
console.time('format')
for (let i = 0; i < 100; i++) {
	actual = format(code, { parser: 'pug' as any, plugins: [plugin] });
}
console.timeEnd('format')
 ```

Here are the times:

With handling of `//- prettier-ignore` comments from this PR: `2685 ms`
With the current master branch: `5223 ms`

So the changes in this PR actually make the processing run twice as fast in this case. This confirms my expectations formulated above, I think.

**Comment by lehni** [comment]
I noticed that the above test is a bit tricky because it's slowed down by logging. If I run with the `--silent` flag, then I don't see the `timeEnd()` measurement. But this ugly trick allows to run silent and still see the times:

```js
const time = Date.now()
for (let i = 0; i < 100; i++) {
	actual = format(code, { parser: 'pug' as any, plugins: [plugin] });
}
expect(Date.now() - time).toBe(0);
```

And with this, I get the following times:

master: `1058ms`
#125: `677ms`

So not quite twice as fast, but 1.56 x faster still.

src/printer.ts
```diff
@@ -733,18 +761,59 @@ export class PugPrinter {
...
-		if (!token.buffer) {
-			result += '-';
-		}
-		result += formatCommentPreserveSpaces(token.val, this.options.commentPreserveSpaces);
-		if (this.nextToken?.type === 'start-pipeless-text') {
-			this.pipelessComment = true;
+		// See if this is a `//- prettier-ignore` comment, which would indicate that the part of the template
+		// that follows should be left unformatted. Support the same format as typescript-eslint is using for descriptons:
+		// https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-ts-comment.md#allow-with-description
+		if (/^ prettier-ignore($|[: ])/.test(commentToken.val)) {
```
Shinigami92, line 769: @ST-DDT Could you take a look at this regular expression and suggest improvements if there are any?

Shinigami92 (approved):

Shinigami92:
@lehni Is something still missing after the conflict has been resolved or can I merge it then?

lehni:
@Shinigami92 No this is all good from my side! I've resolved the conflict also.

PR #132, opened 2020-10-04
Handle for...of in vue v-for

### LINKED ISSUE #129: Bug: Does not work when source contains for...of statement (Vue)
When the pug source code contains a for...of statement, the formatter bails and doesn't do anything in the pug segment.

Replace `v-for="x of quux"` with `v-for="x in quux"` in the input and everything works as intended.

## Info

| Tool     | Version       |
| -------- | ------------- |
| Plugin   | v1.7.0        |
| Prettier | v2.1.0        |
| Node     | v12.18.3        |
| OS       | mac |

## Input

```pug
<!-- Test.vue -->
<template lang="pug">
.test( foo  bar="123" v-for="x of quux")
  | baz {{ quux}}
</template>

<style scoped lang="scss">
.test {
  color:  red;
}
</style>

<script>


export default {
  name: 'Test',
  data: () => ({quux: 'tobi'}),
};
</script>
```

## Actual Output

```vue
<!-- Test.vue -->
<template lang="pug">
.test( foo  bar="123" v-for="x of quux")
  | baz {{ quux}}
</template>

<style scoped lang="scss">
.test {
  color: red;
}
</style>

<script>
export default {
  name: "Test",
  data: () => ({ quux: "tobi" }),
};
</script>
```

## Expected Output

```vue
<!-- Test.vue -->
<template lang="pug">
.test(foo, bar="123", v-for="x of quux")
  | baz {{ quux }}
</template>

<style scoped lang="scss">
.test {
  color: red;
}
</style>

<script>
export default {
  name: "Test",
  data: () => ({ quux: "tobi" }),
};
</script>
```

Shinigami92:
closes #129


Shinigami92:
https://github.com/prettier/plugin-pug/blob/b237d0f05dc4840ef9979a412ada30a4d06a4c85/src/printer.ts#L277

@fisker Hi :slightly_smiling_face: (You are my person of trust in prettier's team ^^ so I write you)
I'm passing the following data to prettier's `format` function
```js
{
  val: 'item of items', // not 'item in items'!
  parser: '__vue_expression',
  codeInterpolationOptions: {
    singleQuote: true,
    bracketSpacing: true,
    arrowParens: 'always',
    printWidth: 9000,
    endOfLine: 'lf'
  }
}
```

Vue seems to **allow** `v-for` with `for of` syntax:  
<a href="https://vuejs.org/v2/guide/list.html">

</a>

Is this a prettier issue? Or do I need to call another (internal) parser for this specific case?

src/printer.ts
```diff
@@ -188,13 +188,17 @@ export class PugPrinter {
...
 						// @ts-expect-error
 						results.push(this[token.type](token));
 						break;
+					}
 				}
-			} catch {
-				throw new Error('Unhandled token: ' + JSON.stringify(token));
+			} catch (error) {
+				throw new Error(error);
 			}
```
lehni, line 202: Why catch just to throw again? You may just not catch at all?

lehni: But it's good this part changes, this used to swallow errors, which has caused some problems for me before also…

src/printer.ts
```diff
@@ -506,7 +510,7 @@ export class PugPrinter {
 				}
 			}
 			if (numNormalAttributes > 0) {
-				// Add leading and trailing parantheses
```
lehni, line 509: Apologies for all these spelling mistakes. It was very late last night and I thought that my VSCode was using a spell checker for this repo 🤦

src/printer.ts
Shinigami92: Yes, just didn't updated this part on a huge rewrite. It was useful some time
I re-throwed it, because in this way I can click on the code line in stack trace and come to this line

fisker:
The `v-for` is similar to `for-of` but not the same, `v-for="(a, b) of c"` is valid, `for ((a, b) of c){}` is not valid js, here is how prettier handle `v-for` https://github.com/prettier/prettier/blob/fbb212ab59a5951f062caf01e94eba89759f24b0/src/language-html/syntax-vue.js#L7

Shinigami92:
@fisker So I did not dig deep into it but the only call to this function is here

https://github.com/prettier/prettier/blob/64cb7abed85566ec5d0eaee13e9d4fc58198f001/src/language-html/printer-html.js#L1081-L1083

```js
  if (options.parser === "vue") {
    if (node.fullName === "v-for") {
      return printVueFor(getValue(), textToDoc);
```

But I'm calling directly with `parser: '__vue_expression'`, so maybe I never reach this code :thinking:  
Maybe I need to call with `parser: 'vue'` directly here in this case
But until know every thing worked with all the internal parsers
https://github.com/prettier/plugin-pug/blob/8106cdae8f5c03466b97300f0c56982729e50e49/src/printer.ts#L368-L413

Shinigami92:
@fisker So indeed if calling

https://github.com/prettier/plugin-pug/blob/a220e916f79f067cf3cba605f237906e18ae9182/src/printer.ts#L396

with only `'vue'` instead of `'__vue_expression'`, the one specific `v-for="x of quux"` test passes!

But why `'__vue_expression'` don't handle this case?
Should we handle it in prettier itself? Or should I add a condition if `v-for` and contains `' of '` then use `parser: 'vue'` :thinking:

Shinigami92:
Awaiting release of prettier: https://github.com/prettier/prettier/pull/9225#issuecomment-710939280

fisker:
> But why '__vue_expression' don't handle this case?

Because it's not legal js code.

Shinigami92:
Oh you mean cause using `of` in js is different to `in`  
But in Vue: `of` is just an alias to `in`

fisker:
No, `of` and `in` the same.

Check https://github.com/prettier/plugin-pug/pull/132#issuecomment-703318209

PR #134, opened 2020-10-04
Improve conversion of tags to literals

### LINKED ISSUE #133: Bug: conversion of literals can replace the wrong content
After having worked a bit with this part of the code, it just occurred to me that there could be this theoretical scenario where the use of `replace()` will replace the wrong part of the content, because it's run on the full result, not just the current element:

## Input

```pug
p
  | Just some text mentioning
  code div#id
  | or perhaps it talks about
  code div.class
div(id="id")
div(class="class")
```

## Expected Output

```pug
p
  | Just some text mentioning
  code div#id
  | or perhaps it talks about
  code div.class
#id
.class
```

## Actual Output

```pug
p
  | Just some text mentioning
  code #id
  | or perhaps it talks about
  code .class
div#id
div.class
```

I have ideas for how to fix this, will work on that next.

lehni:
Closes #133

### DIFF
--- a/src/printer.ts
+++ b/src/printer.ts
@@ -123,6 +123,7 @@ export class PugPrinter {
 	>;
 	/* eslint-enable @typescript-eslint/indent */
 
+	private currentTagPosition: number = 0;
 	private possibleIdPosition: number = 0;
 	private possibleClassPosition: number = 0;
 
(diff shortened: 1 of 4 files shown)


src/printer.ts
```diff
@@ -547,11 +564,8 @@ export class PugPrinter {
 			if (isQuoted(token.val)) {
 				if (token.name === 'class') {
 					// Handle class attribute
-					let val: string = token.val;
-					val = val.slice(1, -1);
-					val = val.trim();
-					val = val.replace(/\s\s+/g, ' ');
-					const classes: string[] = val.split(' ');
```
lehni, line 554: When looking for use-cases `replace()`, I found these lines and realized they could be optimized.

lehni:
@Shinigami92 this change should actually speed things up a little, since `replace()` is now only called on the tag string, not the full content in `this.result`

src/printer.ts
```diff
@@ -264,6 +265,21 @@ export class PugPrinter {
 		return parts;
 	}
 
+	private maybeReplaceTagWithLiteral(search: RegExp, replace: string): void {
```
Shinigami92, line 268: I (personally) don't like functions having `maybe` in it
_Will it do it or will it not?_ :man_shrugging: 
I wan't to be sure when calling a function ^^

Shinigami92 (changes requested):

src/printer.ts
lehni: Well the thing is, it doesn't know if it will or not until it tries… That's how the code so far worked also.

Shinigami92: I didn't criticize the way it works but the name of the function :slightly_smiling_face:  

Let's go with `replaceTagWithLiteralIfPossible`, IMO that's an okay-ish name

lehni: Haha no I understand! I wasn't very happy about it either, but couldn't think of a better name at the time. I actually think this whole literal handling could maybe be done in a different way that isn't relying on replacing strings in `this.results` (feels a bit hackish?), and have some thoughts about it already, but first I wanted to fix the actual bugs :)

lehni: BTW: I'm really enjoying the exchanges here. It's so much more enjoyable than what is usually the case when working on open-source. 👍

Shinigami92 (approved):

PR #153, opened 2020-11-01
Add option for use class and id literals

### LINKED ISSUE #36: Option for disabling formatting of Class and ID literals
Let the user decide if the formatter should format Class and ID literals or not

---

based on #32

SkyaTura:
Closes #36 

## Todo
- [x] Add tests
- [x] Option to preferred literals
- [x] Add docs
- [ ] ~Option to preferred attributes~ _this PR won't have this use case_


SkyaTura:
I don't know yet how to properly add new attributes in case there is none in the line.

🙈

src/options/use-class-literals.ts
```diff
@@ -0,0 +1,12 @@
+import { BooleanSupportOption } from 'prettier';
+import { CATEGORY_PUG } from '.';
+
+export const PUG_USE_CLASS_LITERALS: BooleanSupportOption = {
+	since: '1.10.0',
```
Shinigami92, line 5: `1.10.0` was already released and I will release `1.10.1` today as a bugfix version
This feature will potentially land in `1.11.0` (if you manage to handle this PR :wink:)

SkyaTura: I leave the release version because I don't know when it will be released :P

src/options/use-class-literals.ts
```diff
@@ -0,0 +1,12 @@
...
+	category: CATEGORY_PUG,
+	type: 'boolean',
```
Shinigami92, line 7: As discussed in the issue, we should rewrite this to a `choice` (same for _id literal_)

Shinigami92 (changes requested):

src/options/use-class-literals.ts
SkyaTura: I was just doing that 😂

Shinigami92:
Will update deps in master, give me one second :slightly_smiling_face:

SkyaTura:
> Will update deps in master, give me one second 🙂

ops, I'll revert my last commit

Shinigami92:
There it is :slightly_smiling_face:

src/options/converge.ts
```diff
@@ -31,6 +31,8 @@ export function convergeOptions(options: ParserOptions & PugParserOptions): PugP
 		pugSortAttributesEnd: options.pugSortAttributesEnd,
 		pugWrapAttributesThreshold: options.pugWrapAttributesThreshold,
 		pugWrapAttributesPattern: options.pugWrapAttributesPattern,
+		pugIdNotation: options.pugIdNotation,
+		pugClassNotation: options.pugClassNotation,
```
Shinigami92, line 35: mh, just a minor thing, but everywhere it seams we have class then id
so we could switch these two line ^^'

Shinigami92 (changes requested):
Could you also add a pug test-case for each case were a line will wrap?

Shinigami92:
After these two changes, I think I will merge it for now and then switch to `master`->`main`  
Should we then create a 1.13.0 release? Or should we try to also support the third option?

SkyaTura:
I think these two options already have it's values, and the third option (afaik) seems complicated to me, so it may take some time to implement it.

README.md
```diff
@@ -257,6 +257,26 @@ These additional options are specific to pug templates and can be configured in
...
   Define a list of patterns for attributes that will be forced to have empty quotes even with "none" selected.
 
+- `pugClassNotation`  
+  Define how classes should be formatted.
+
+  Choices:
+
+  - `'as-is'` -> Disables class formatting.  
+    Example: `foo.bar(class="baz")`
+  - `'literal'` _default_ -> Forces all valid classes to be printed as literals.
```
Shinigami92, line 267: Could you swap these options, so the default will placed first
Also make a comment, that the third option `attributes` will come later (could we style this in any way? like grey-ish text-color?)

src/options/pug-class-notation.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const PUG_CLASS_NOTATION: ChoiceSupportOption = {
+	since: '1.13.0',
+	category: CATEGORY_PUG,
+	type: 'choice',
+	default: 'literal',
+	description: 'Define how classes should be formatted.',
+	choices: [
+		{ value: 'as-is', description: 'Disables class formatting.' },
+		{ value: 'literal', description: 'Forces all valid classes to be printed as literals.' }
```
Shinigami92, line 12: Also we could switch these choices
I think this affects the order when using the CLI `--help` (not that anyone uses it 😅 but hey, doing it directly the right way 😁)

SkyaTura: should I also switch on the readme?

Shinigami92: Yeah, everywhere
So it will ever be
1. literal
2. attribute
3. as-is

Shinigami92:
Okay 😀
Should I merge it? Or do you want to do anything else from your side?

The third attribute can be places now here in grey-ish or later in the other PR when it's implemented, I will leave this to you.

Shinigami92 (approved):

PR #159, opened 2020-11-23
Format inline scripts and styles

### LINKED ISSUE #158: Bug: Inline JavaScript in text block is not pretty
It seems that Pug `script.` blocks aren't parsed / printed properly.  It should make the JavaScript code prettier.

## Info

| Plugin    | v1.10.1                  |
| Prettier  | v2.1.2                  |
| Framework | none |
| Node      | v14.15.0                  |
| OS        | linux           |

## Prettier config

```json
{
	"useTabs": true,
	"semi": false,
	"arrowParens": "avoid",
	"trailingComma": "es5",
	"overrides": [{
		"files": "*.riot",
		"options": {"parser": "pug"}
	}]
}
```

Just ignore what I'm doing with *.riot file extensions.  This bug exists even when `--no-config` flag is passed to prettier.

## Input

```pug
script.
	const   x   =   2   ;   // comment
```

## Output or Error

```bash
cat test.pug | npx prettier --parser pug
```

```pug
script.
	const   x   =   2   ;   // comment
```

## Expected Output

```pug
script.
	const x = 2 // comment
```

## Compare this to Prettier's HTML parser:

```
cat | npx prettier --parser html
<script>
	const   x   =   2   ;   // comment
</script>
```

which outputs

```html
<script>
	const x = 2 // comment
</script>
```

Notice missing semicolon and proper spacing

Shinigami92:
closes #158

### DIFF
--- a/src/printer.ts
+++ b/src/printer.ts
@@ -64,6 +64,7 @@ import {
 	isQuoted,
 	makeString,
 	previousNormalAttributeToken,
+	previousScriptTagToken,
 	unwrapLineFeeds
 } from './utils/common';
 import { isVueEventBinding, isVueExpression, isVueVForWithOf } from './utils/vue';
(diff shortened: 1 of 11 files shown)


src/printer.ts
```diff
@@ -1056,7 +1057,63 @@ export class PugPrinter {
...
+					index++;
+					tok = this.tokens[index];
+				}
+
+				result = format(rawText, { parser, ...this.codeInterpolationOptions });
+				const indentString: string = this.indentString.repeat(this.indentLevel + 1);
+				result = `\n${result
+					.split('\n')
+					.map((line) => indentString + line)
+					.join('\n')}`.trimRight();
```
ST-DDT, line 1104:
```suggestion
				result = `\n${result
				    .trimRight()
					.split('\n')
					.map((line) => indentString + line)
					.join('\n')}`;
```
If we remove the newline first, then we don't have to prefix it with the indentation before removing it (minimal performance gain).

ST-DDT (approved):

src/printer.ts
Shinigami92: Ah okay, but I would prefer to write:
```ts
result = format(rawText, { parser, ...this.codeInterpolationOptions });
result = result.trimRight();
const indentString: string = this.indentString.repeat(this.indentLevel + 1);
result = result.split('\n').map((line) => indentString + line).join('\n');
result = `\n${result}`;
```
Should be the same, but much more readable, right?

ST-DDT (approved):

PR #163, opened 2020-11-25
Format value of style attribute in tags

### LINKED ISSUE #162: Format value of style attribute in tags
## Info

| Plugin    | v1.11.0                 |
| Prettier  | v2.2.0                  |
| Framework | none |
| Node      | v15.2.1                  |
| OS        | win,linux,mac           |

## Prettier config

```json
{}
```

## Input

```pug
span(style="color:red;background-color:yellow")
```

## Output or Error

Same a input

## Expected Output

```pug
span(style="color: red; background-color: yellow")
```

Shinigami92:
closes #162


ST-DDT:
Have you tried it with something like this: https://www.w3schools.com/cssref/pr_gen_content.asp
````css
a::after {
  content: " (Link:" attr(href) ";)";
}
````
Especially if it needs to be escaped for usage inside inline attributes. I didn't manage to to embed it into an inline css though.

However, I managed to use this in plain html
````html
<q style='quotes: "«" "»" "‹" "›";'>This is a <q>nested</q> quote.</q>
````
Or just to annoy any parser:
````html
<q style='quotes: ";" ";" ";" ";";'>This is a <q>nested</q> quote.</q>
````

ST-DDT:
Or something like this:
````html
<q style="quotes: &quot;;&quot; &quot;;&quot; &quot;'&quot; &quot;'&quot;;">This is a <q>nested</q> quote.</q>
````

Shinigami92:
@ST-DDT Thank you :+1: 

As you figured out yourself, I think your first example isn't covered by this PR, because this is only for the style attribute of a tag and not the style-tag by itself

But the other cases are interesting examples and I will try to add them and find out what happens
I would expect that the prettier css parser will going wild with it and throw an error
Then I maybe need to handle that or it's a valid error that will come up in the users terminal

Remember that my formatter is just a formatter and not a parser or linter :smile:

Shinigami92:
Ok, so `span(style="quotes: &quot;;&quot; &quot;;&quot; &quot;'&quot; &quot;'&quot;;")` will be rejected by the prettier parser

But yes you were right and I needed to fix the quotation types

Issue #58, opened 2020-01-08
Handle Attribute Interpolation not using template literals and using quotes


Shinigami92:
https://pugjs.org/language/attributes.html#attribute-interpolation

Merged Issues #53. #54 and #55 into this

Shinigami92:
@lehni Do you also want to work on some of these `help wanted` issues?

Shinigami92:
@leider @Cloudless67 This is now released as [v1.13.4](https://www.npmjs.com/package/@prettier/plugin-pug/v/1.13.4) :tada:  
Thank you very much :heart:

PR #179, opened 2021-02-18
Fix attribute interpolation using quotes

Cloudless67:
This is fix for issue #58.

### DIFF
--- a/tests/issues/issue-58/issue-58.test.ts
+++ b/tests/issues/issue-58/issue-58.test.ts
@@ -1,0 +1,14 @@
+import { readFileSync } from 'fs';
+import { resolve } from 'path';
+import { format } from 'prettier';
+import { plugin } from './../../../src/index';
+
+describe('Issues', () => {
+	test('attribute interpolation using quotes', () => {
+		const expected: string = readFileSync(resolve(__dirname, 'formatted.pug'), 'utf8');
+		const code: string = readFileSync(resolve(__dirname, 'unformatted.pug'), 'utf8');
+		const actual: string = format(code, { parser: 'pug', plugins: [plugin] });
+
+		expect(actual).toBe(expected);
+	});
+});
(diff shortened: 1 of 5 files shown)


tests/issues/issue-58/issue-58.test.ts
```diff
@@ -0,0 +1,14 @@
+import { readFileSync } from 'fs';
+import { resolve } from 'path';
+import { format } from 'prettier';
+import { plugin } from './../../../src/index';
+
+describe('Issues', () => {
+	test('attribute interpolation using quotes', () => {
```
Shinigami92, line 7: I try to use sentences like "it should ..."
sadly jest uses the `test` function instead of `it`
Could you try to reformulate the sentence to e.g. `should interpolate attribute using quotes`? :slightly_smiling_face:

Shinigami92 (changes requested):
It seems to work :tada:  
Really much :heart: to you for fixing this really old issue :smiley:  
I think you will make many devs in our community happy with that!

One more thing you can do is removing all the parts that addressing this workaround like in the `README.md`

I will update the changelog on my own and managing the next version number, so you don't need to check for these.

Shinigami92:
Oh, and one more thing that would be cool: Could you write inline comments or improve the JSDoc for the function to address this improved special behavior?  
It's not really clear what the regex pattern is doing and why it is doing so. :smile:

Cloudless67:
First, thanks for quick review:)

I'll try to work on things you've said!

Shinigami92:
@Cloudless67 I'm currently in some meetings and think I can create a release todays evening

A little minor change you could make is correcting the capital typo in the comment
```
Regex for Checking if there are any unescaped quotations.
Regex for checking if there are any unescaped quotations.
```

Shinigami92 (approved):

tests/issues/issue-58/issue-58.test.ts
JounQin: @Shinigami92 AFAIK, `it` is also available in jest.

Shinigami92: Yeah, but only as an alias (https://jestjs.io/docs/en/api#testname-fn-timeout)
I think the only reason was to stick with non-aliased / native jest functions so users can find documentation more quickly and are not confused because they don't know other test tools like mocha

PR #183, opened 2021-03-19
Handle autocloseable tag with trailing text correctly

### LINKED ISSUE #182: Bug: Treat the autocloseable tag with trailing text correctly
## Info

| Plugin    | v1.13.5                 |
| Prettier  | v2.2.1                  |
| Framework | none                    |
| Node      | v15.6.0                 |
| OS        | win,linux,mac           |

## Input

```pug
span Text
  br/ more text
```

## Expected Output

```pug
span Text
  br/ more text
```

## Actual Output

```pug
span Text
  br/more text
```

Shinigami92:
closes #182


Shinigami92:
Hi @ST-DDT o/ :smile:  
Do you have any additional suggestions? Maybe some more test cases?
Or anything I can improve in the code itself?

ST-DDT:
I'm not sure about the test, because it only tests that it stays as it is, but maybe there is nothing else to be tested there.
One thing you might want to test, is text that is composed of paragraphs/linebreaks and continuous sections. Similar to the following html section:

````html
<span>This is a very long sentence so it got
    wrapped automatically at the end of the line.
    <br />
    But some things need to be highlighted beyond the capabilties of a
    single newline, so lets try to use two too:
    <br />
    <br />
    Foobar is the the result of naming a bar foo!
</span>
````

I will contact you in person to discuss this PR in detail later.

---------

PS: Please fix the build to show warnings like these only once:

> You have a misspelled word: autocloseable on String

**Comment by Shinigami92** [comment]
I think your example would look like 

```pug
span This is a very long sentence so it got
  | wrapped automatically at the end of the line.
  br/
  | But some things need to be highlighted beyond the capabilities of a
  | single newline, so lets try to use two too:
  br/
  br/
  | Foobar is the result of naming a bar foo!
```

You are indicating in the first sentence that you predict that the line would wrap by my formatter.
This is not the case!
I may consider to implement such a mechanic, but that would be a new issue.
Also it could be really hard, because I don't have to only check if the length is to long, but also if the line can hold more content of the further line and so on. :thinking: 

---

The warnings in the `Files changes` tab are coming from the linter step in the GitHub Actions workflow.
I could add an `if` to perform linting only on `linux` + `node-lts` :thinking:  
We should move this into a new issue if it really worth the effort

ST-DDT (approved):
LGTM 👍

Shinigami92:
> Please fix the build to show warnings like these only once

done :heavy_check_mark:

Issue #172, opened 2020-12-18
Disable PugPrinter:formatText vue's specific warnings


bsiguret:
## Request / Idea

Make a way to be able to disable these kind of warning:

`If you are using Vue, you can ignore this message`

## Additional Context
Hello,

First of all sorry if it's existing already but I couldn't find a way to do this. We are working on Vue with pug template and Nuxt. Our problem is that Pug warning are spamming our terminal every time our site hot reload with this type of messages:

`[PugPrinter:formatText]: Missing expected ). If you are using Vue, you can ignore this message. code:`

I thought there would be an option to disable these vue's specifics warning but I couldn't find it

Shinigami92:
There is one problem with that: It's more "you may ignore this message" instead of "can"
What if this message was not printed, it it's actually a problem :thinking: 

We already thought about flags for vue and angular, but until now we tried to avoid these flags

You should add a comment on #109 (it's now in GitHub's new `Discussion` section)

_little sidenote: I'm currently making a bit like holiday from working on GitHub, just responding here and there. Will effectively work again starting January 2021_

bsiguret:
Yeah, my point is to have a flag for Vue to disable these warning who aren't an actual problem, because vue.

It wouldn't be a problem if they were printed only one time but because of Nuxt hot reload they are literally spamming and polluting our terminal. It may be a specific problem thought. I'll check the Discussion and add a little comment linking to this issue.

_Aren't we all ? Thanks for the quick response and enjoy your holidays_

Shinigami92:
@bsiguret 

Thinking about this futher, should I just implement a `pugLoggerLevel` option?  
So users can set it to `pugLoggerLevel: 'off'` if they really want to?

The default would still be 'info', but this way you could set it to e.g. 'error'.
Downside would be you will ignore ANY warnings.
But I don't really want to provide something like ignore for a framework or ignore error x and y and z.
So I don't have error-codes and I also really don't want them cause I think it would be overkill.

What do you think? What would be the at least okay-ish version for you to be satisfied?

bsiguret:
@Shinigami92 

Hello, Sorry for the delay. I don't think it is a good solution to off all warnings. Some might be very useful.

Shinigami92:
@bsiguret Thank you.
This is helpful feedback and I will take this into account when designing a suitable option.

broofa:
[Leaving a breadcrumb here for anyone else that stumbles across this.]

[...]

PR #186, opened 2021-03-27
Add framework option

### LINKED ISSUE #184: Bug: Interpolating Javascript codes creates a warn
## Info

| Plugin    | v1.13.5                  |
| Prettier  | v2.x.x                  |
| Framework | vue |
| Node      | v14.15.3                  |
| OS        | mac           |

## Prettier config

```json
{}
```

## Input

```pug
p {{ new Date().toISOString() }}
```

## Output or Error

```bash
console.warn
    [PugPrinter:formatText]:  SyntaxError: Unexpected token 'Date'
        at o (/Users/skyatura/projects/shinigami92/plugin-pug/node_modules/prettier/parser-angular.js:46:17262)
        at Object.<anonymous>.e.parseNgInterpolation (/Users/skyatura/projects/shinigami92/plugin-pug/node_modules/prettier/parser-angular.js:46:18429)
        at s (/Users/skyatura/projects/shinigami92/plugin-pug/node_modules/prettier/parser-angular.js:46:29190)
        at Object.<anonymous>.e.parseInterpolation (/Users/skyatura/projects/shinigami92/plugin-pug/node_modules/prettier/parser-angular.js:46:29490)
        at /Users/skyatura/projects/shinigami92/plugin-pug/node_modules/prettier/parser-angular.js:46:30113
        at Object.parse (/Users/skyatura/projects/shinigami92/plugin-pug/node_modules/prettier/parser-angular.js:46:29796)
        at Object.parse (/Users/skyatura/projects/shinigami92/plugin-pug/node_modules/prettier/index.js:13625:19)
        at coreFormat (/Users/skyatura/projects/shinigami92/plugin-pug/node_modules/prettier/index.js:14899:14)
        at format (/Users/skyatura/projects/shinigami92/plugin-pug/node_modules/prettier/index.js:15131:14)
        at /Users/skyatura/projects/shinigami92/plugin-pug/node_modules/prettier/index.js:57542:12
```

SkyaTura:
This PR adds a new option that allows users to specify which framework is being used in the project.

This is needed to fix #184 

TODO:

- [x] Use `babel` as default parser
- [x] Add Svelte (although there is no specificity to this framework ~yet~)
- [x] Add documentation
- [x] Add framework detection
- [x] Test framework detection
- [ ] Test framework specific warnings


SkyaTura:
@Shinigami92 could you please tell me if it's ok to create this option. 😄

Shinigami92:
> @Shinigami92 could you please tell me if it's ok to create this option.

image

Shinigami92:
I will make a more in depth review tomorrow :+1:

Shinigami92:
I'm currently not sure if there is something special that needs to be done for `svelte` (never used it on my own)
But it could have similar syntax to Vue, I think

So what do you think about using `babel` as the default parser and only use `__ng_interpolation` if it is actually `angular`?
This would be a breaking change, but this may be okay
But the breaking change would be more _okay_ if we also directly solve #172
Do you think you could tackle this also in this PR? May only be "_if vue, then don't log_" for now

SkyaTura:
I was thinking on avoid breaking changes for now, since it works fine (except for the warnings).

I'll ignore the vue's specific errors to solve #172

Shinigami92:
I try to get in contact with a pug-svelte developer currently
https://discord.com/channels/457912077277855764/750401468569354431/825717140195639367

image

---

I'm just _afraid_ using the `__ng_interpolation` as the default for `'none'`, cause in theory it doesn't make sense in any way.
The `__ng_*` parsers are specially internal angular parsers, and it is a good thing if we can get rid as most as possible of them.

Don't fear a breaking change here, it is a `1.4.0` and I will later explicitly make a note of this in the changelog.

---

Beside that, I think I will create a [VuePress](https://vuepress.vuejs.org) (or similar) documentation.

Shinigami92:
Just to get not confused ^^
You are going to solve the TODOs in this PR, will you?

I will convert this PR to a draft, just convert it back when you are finished and I can review the docs :slightly_smiling_face:

SkyaTura:
Yes, I will as soon as I get some time to 😄

SkyaTura:
@Shinigami92 how do I test warnings? 🤔

Shinigami92:
> How do I test warnings? 

I thought about this a long time ago.
Short: currently there are no tests for doing that.
But IMO we should think about a test-strategy (maybe not in this issue, but in a separat one)

> Also, when you merge #191 will we maintain the README.md documentarion?

You should currently just add a documentation to the actual `main` `README.md`
I will keep track of that, when this issue is merged to `main`
The documentation website may take some more time and it's heavily in development :slightly_smiling_face: 

But yes, later we don't need to add the documentation of options in the `README.md` but the website
There is a script `yarn docs:dev` to spin up the docs on `localhost:8080`

So for now you should add a row between line `144` (`pugWrapAttributesPattern`) and `145` (`pugSingleFileComponentIndentation`), cause I think it fits best above `pugSingleFileComponentIndentation` :slightly_smiling_face:  
You may use `don't wrap lines` in your IDE to add it in there :smile:

src/options/pug-framework.ts
```diff
@@ -0,0 +1,20 @@
...
+export const PUG_FRAMEWORK: ChoiceSupportOption = {
+	since: '1.14.0',
+	category: CATEGORY_PUG,
+	type: 'choice',
+	default: 'none',
+	description: 'Define which framework is used in the project.',
+	choices: [
+		{ value: 'none', description: 'No framework specifically.' },
+		{ value: 'angular', description: 'Uses Angular.' },
+		{ value: 'svelte', description: 'Uses Svelte' },
```
Shinigami92, line 14:
```suggestion
		{ value: 'svelte', description: 'Uses Svelte.' },
```

src/printer.ts
```diff
@@ -415,20 +433,26 @@ export class PugPrinter {
...
-								logger.warn(
-									'[PugPrinter:formatText]: Missing expected ). If you are using Vue, you can ignore this message.',
-									`code: \`${code.trim()}\``
-								);
-							} else if (error.includes('Missing expected :')) {
-								logger.warn(
-									'[PugPrinter:formatText]: Missing expected :. If you are using Vue, you can ignore this message.',
-									`code: \`${code.trim()}\``
-								);
+								if (pugFramework !== 'vue') {
```
Shinigami92, line 436: **thought:** Mh :thinking:  
We should really find out how `svelte` behaves here :eyes: (and the below ones)

Cause maybe we need to use `pugFramework === 'angular'` here
But I'm absolutely not sure

tests/issues/issue-70/issue-70.test.ts
```diff
@@ -7,7 +7,12 @@ describe('Issues', () => {
 	test('should handle Framework interpolation', () => {
```
Shinigami92, line 7:
```suggestion
	test('should handle angular framework interpolation', () => {
```

tests/issues/issue-9/issue-9.test.ts
```diff
@@ -11,6 +11,8 @@ describe('Issues', () => {
 			parser: 'pug',
 			plugins: [plugin],
 
+			// @ts-expect-error
+			pugFramework: 'angular',
```
Shinigami92, line 15: _cannot add comment to line 7 ^^_

`should escape quotes correctly in angular`

Shinigami92:
@SkyaTura Just an idea, but do you think it is possible to detect which framework is in use?
e.g. read the projects package.json an look for a dependency of `angular` or `vue` or `svelte`?

And then use this as fallback if no `pugFramework` is provided?

I think we cannot write tests for that (at least not easily) so we need to just test it in a local copy/repo.
I'm thinking about using `fs` calls :thinking:

Shinigami92:
@SkyaTura I checked how to possibly detect the projects framework.
You may read e.g. `process.env.npm_package_dependencies_vue`. The value can be `undefined` or (in my case) `'~2.6.12'`.
I think other dependencies can be checked exactly the same way :slightly_smiling_face: 

So we may use something like

```js
if (process.env.npm_package_dependencies_vue)
  pugFramework = 'vue'
else if (process.env.npm_package_dependencies_svelte)
  pugFramework = 'svelte'
else if (process.env.npm_package_dependencies_angular)
  pugFramework = 'angular'
else
  pugFramework = 'none'
```

src/printer.ts
Shinigami92: We'll see if some people run into this problem
If so, they will create an issue

Shinigami92:
@SkyaTura I'm not sure how and if we should test framework specific warnings :thinking:  
Nevertheless I worked on all other remaining tasks :smiley:  
Could you have a look and review my changes? _I know it's your branch, so you cannot give your approval. So just write a comment to let me know_

SkyaTura:
It seems fine to me, thank you for helping. 😸

Shinigami92 (approved):

Issue #167, opened 2020-11-28
Add third option for pugClassNotation and pugIdNotation


Shinigami92:
This will be a follow-up of #36, because this third option will need a bit more refactoring and is not quite as easy to implement as the other two options.  
To let users already benefit from the other two options, we will create a release before starting working on this third option.

h0merjam:
This will be extremely helpful when using Tailwind!

Has this been started? Are there any pointers available for where to begin?

Shinigami92:
It's kinda easy to start by adding the new option to the options:

https://github.com/prettier/plugin-pug/blob/a66432dc9890a52c671ac77f7eb58a4f1f7e241c/src/options/pug-class-notation.ts#L11-L14

https://github.com/prettier/plugin-pug/blob/a66432dc9890a52c671ac77f7eb58a4f1f7e241c/src/options/pug-id-notation.ts#L11-L14

And then add tests for the option in

- https://github.com/prettier/plugin-pug/tree/main/tests/options/pugClassNotation
- https://github.com/prettier/plugin-pug/tree/main/tests/options/pugIdNotation

From there you can then just try-and-error through the code with test-driven-development ^^

Just use `yarn test tests/options/pugClassNotation` and `yarn test tests/options/pugIdNotation`

Shinigami92:
@h0merjam The third choice should be `attribute` (opposite of `literal` ^^)

And then your both function friends are here and below: https://github.com/prettier/plugin-pug/blob/main/src/printer.ts#L577

`start-attributes` and `attribute`

h0merjam:
Hmm, this isn't quite as simple as I'd hoped!

I've tried to grasp what's happening with the `literal` option and how the `attribute` option might work. Based on what I think I understand so far the steps might be:

- store literals via `class`/`id` token type processor (this would require adding new private vars to the `PugPrinter` class❓ - `classesToAppend` / `idToAppend`)
- convert/append literals to `class`/`id` attribute (either in existing class attribute or `end-attributes` / `newline` hooks)

Does this sound about right? 🤔

Shinigami92:
I'm currently not at home 
But yes that sounds right 
Try the one that is easier, in the end both could work

PR #203, opened 2021-04-21
Add pugClassNotation 'attribute' option

h0merjam:
This is a partial fix for #167 covering classes, I think it _should_ be fairly straightforward to implement for IDs next, please let me know if this direction is right and I'll get on to it!

Thanks!

### DIFF
--- a/src/printer.ts
+++ b/src/printer.ts
@@ -186,6 +186,8 @@ export class PugPrinter {
 	private pipelessComment: boolean = false;
 	private currentlyInPugInterpolation: boolean = false;
 
+	private readonly classLiteralToAttribute: string[] = [];
+
 	/**
 	 * Constructs a new pug printer.
 	 *
@@ -689,7 +691,7 @@ export class PugPrinter {
 
 		if (typeof token.val === 'string') {
 			if (isQuoted(token.val)) {
-				if (token.name === 'class' && this.options.pugClassNotation !== 'as-is') {
+				if (token.name === 'class' && this.options.pugClassNotation === 'literal') {
 					// Handle class attribute
 					const val: string = token.val.slice(1, -1).trim();
 					const classes: string[] = val.split(/\s+/);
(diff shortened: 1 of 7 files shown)


src/printer.ts
```diff
@@ -773,6 +775,20 @@ export class PugPrinter {
 			if (token.val !== true) {
 				this.result += `=${token.val}`;
 			}
+		} else if (token.name === 'class' && this.options.pugClassNotation === 'attribute') {
+			const val: string = token.val.slice(1, -1).trim();
```
Shinigami92, line 779: **thought:** Havn't tested the code yet, and I'm currently just in GitHub Web UI as writing this review.
But this looks dangerous, cause the `token.val` could be not quoted and include js code 🤔, could it?

src/printer.ts
```diff
@@ -838,6 +854,24 @@ export class PugPrinter {
 			this.result += this.indentString.repeat(this.indentLevel);
 		}
 		this.wrapAttributes = false;
+
+		if (this.classLiteralToAttribute.length) {
+			if (this.previousToken?.type === 'start-attributes') {
+				this.result += '(';
+			}
+
+			if (this.previousToken?.type === 'attribute') {
```
Shinigami92, line 863: **suggestion (performance):** Using at least an `else if` here could safe one check, cause when it was already a `start-attribute` it cannot be an `attribute` anymore

src/printer.ts
Shinigami92: You should add a test case for this 🙂

Shinigami92 (commented):
Code is looking good until now 👍 
You can proceed like this
Do you want to handle `id` in another PR? Or currently just solve it for `class`?

h0merjam:
You know what, a separate PR would be better. It's classes I really need and I'm not sure I'll get on to IDs soon. Hopefully this helps anyone else wanting to have a go at it!

Shinigami92:
Could you update https://github.com/prettier/plugin-pug/blob/main/docs/guide/pug-specific-options.md
and add a test case that contains class with js code

e.g.

```pug
- var clazz = 'ma-2 px-1'
div(class=clazz)
```

Shinigami92:
Ah, these whitespaces are necessary cause otherwise VuePress will render these in one line instead of adding a newline

h0merjam:
Oh no! I have an overzealous markdown formatter I'm afraid. Whitespace is restored now...

src/printer.ts
```diff
@@ -773,6 +775,20 @@ export class PugPrinter {
 			if (token.val !== true) {
 				this.result += `=${token.val}`;
 			}
+		} else if (token.name === 'class' && this.options.pugClassNotation === 'attribute') {
+			const val: string = isQuoted(token.val) ? token.val.slice(1, -1).trim() : token.val;
+			const classes: string[] = val.split(/\s+/);
```
Shinigami92, line 780: **thought:** I'm a little bit afraid of this line and don't know what it's doing if the class contains js code with spaces 🤔  
But I think I will merge it for now

Shinigami92 (approved):
Cause of work for my employer, I hope I can setup and bundle everything together this evening or over the weekend

PR #204, opened 2021-04-22
Handle svelte syntax

### LINKED ISSUE #201: Bug: svelte pug formatting
## Info

| Plugin    | v1.13.5                  |
| Prettier  | v2.2.1                  |
| Framework | svelte |
| Node      | v15.4                  |
| OS        | wsl           |

## Additional Context

In reference to [https://github.com/sveltejs/language-tools/issues/951](https://github.com/sveltejs/language-tools/issues/951) please find the basic repository below.

[https://github.com/basaran/svelte-pug](https://github.com/basaran/svelte-pug)

Additional information has been provided in the readme.md for setting up the environment.

Shinigami92:
close #201 

This also affected the behavior of `interpolation-token` tokens and it's bracket spacing handling


src/printer.ts
```diff
@@ -487,8 +489,67 @@ export class PugPrinter {
...
+									code = code.slice(1);
+								}
+							} catch (error: unknown) {
+								logger.warn(error);
+							}
+						}
+						code = unwrapLineFeeds(code);
+						result += handleBracketSpacing(this.options.pugBracketSpacing, code, ['{', '}']);
+						text = text.slice(end2 + 1);
+					} else {
```
Shinigami92, line 544: This section is just a stupid copy-paste of the above code, but it checks for one instead of two curly brackets...  
We should refactor it and find out if there is a better solution.
Also the error handling was removed except a default warning and it's currently only checking if framework is set to `svelte`

We also need more test cases with text svelte interpolation. Currently I observed that the whole part in the `try` block is failing and always the `catch` fallback algorithm is taken :thinking:.
Therefore can it removed for svelte, or are there possibilities where it's useful?

src/printer.ts
```diff
@@ -550,6 +611,23 @@ export class PugPrinter {
 		return this.quoteString(val);
 	}
 
+	private formatSvelteInterpolation(val: string): string {
```
Shinigami92, line 614: This function is a copy of `formatAngularInterpolation` but with checking for single curly brackets.
Maybe we can merge them and make it more dynamic. Or maybe we need to find out if it's useful for other cases then only `svelte` :thinking:

src/utils/common.ts
```diff
@@ -152,10 +152,13 @@ export function isMultilineInterpolation(val: string): boolean {
...
  * @param bracketSpacing Specifies whether or not to insert spaces before and after the code.
  * @param code Code that is enclosed in brackets.
+ * @param param2 Brackets.
+ * @param param2."0" Opening brackets.
+ * @param param2."1" Closing brackets.
  * @returns The handled string.
  */
-export function handleBracketSpacing(bracketSpacing: boolean, code: string): string {
-	return bracketSpacing ? `{{ ${code} }}` : `{{${code}}}`;
+export function handleBracketSpacing(bracketSpacing: boolean, code: string, [opening, closing] = ['{{', '}}']): string {
```
Shinigami92, line 160: Maybe we should use another kind of parameter like `1` or `2` :thinking:  
But this could result in more processing
Maybe a benchmark could help here

src/utils/svelte.ts
```diff
@@ -0,0 +1,26 @@
...
+ * ```
+ *
+ * In this case `val` is `"{ cat.id }"`.
+ *
+ * ---
+ *
+ * @param val Value of tag attribute.
+ * @returns `true` if `val` passes the svelte interpolation check, otherwise `false`.
+ */
+export function isSvelteInterpolation(val: string): boolean {
```
Shinigami92, line 18: This is a copy of `isAngularInterpolation` but with only checking for single curly brackets

src/printer.ts
```diff
@@ -1130,7 +1210,7 @@ export class PugPrinter {
 				break;
 		}
 		result += token.mustEscape ? '#' : '!';
-		result += `{${token.val}}`;
+		result += handleBracketSpacing(this.options.pugBracketSpacing, token.val.trim(), ['{', '}']);
```
Shinigami92, line 1213: I accidentally found out that we could support bracket spacing for `interpolated-code` tokens :astonished:  
Maybe we should create it's own PR so we can squash-merge this and have a better history in `main` branch

Shinigami92:
@GrygrFlzr @basaran @dummdidumm
hey :smiley: 
Could you have a look over this PR?
This a first working solution for `svelte` :tada: 
Maybe someone of you could try it out in a production environment? With `yarn link` or `patch-package`?

src/printer.ts
Shinigami92: Extracted this to its own PR in #208

src/printer.ts
```diff
@@ -393,6 +394,7 @@ export class PugPrinter {
 	private formatText(text: string): string {
 		let result: string = '';
 		while (text) {
+			// Find double curly brackets
```
ST-DDT, line 397: Refactor this entire method into smaller parts (that can be reused and are easier to grasp)

src/printer.ts
ST-DDT: IMO you should merge these two methods into a single one and use the existing methods to call that method with the right parameters.

- `private formatAngularInterpolation(val: string): string {`
- `private formatSvelteInterpolation(val: string): string {`
- `private formatFrameworkInterpolation(val: string, quotes: string[]): string {` (not sure about the name 😉 )

src/utils/svelte.ts
```diff
@@ -0,0 +1,26 @@
...
+	return (
+		val.length >= 3 &&
+		((val[0] === '"' && val[val.length - 1] === '"') || (val[0] === "'" && val[val.length - 1] === "'")) &&
```
ST-DDT, line 21: These expressions should be simplified into a separate method calls as well.

````js
return (
    val.length >= 3 &&
    (isWrappedWith(val, '"') || isWrappedWith(val, "'")) &&
    isWrappedWith(val, '{', '}', 1) &&
    !val.includes('{', 2)
);
````


````js
function isWrappedWith(val: string, wrapper: string, offset: int = 0) {
	return function isWrappedWith(val, wrapper, wrapper, offset);
}

function isWrappedWith(val: string, start: string, end: string, offset: int = 0) {
    return val[offset] === start && val[val.length - 1 - offset] === end;
}
````

**Review by ST-DDT — CHANGES_REQUESTED** [review]

File: src/utils/svelte.ts (hunk 7 above)
Comment (ST-DDT) [reply src/utils/svelte.ts:21]: Maybe you can even add something like this:

````js
function isQuoted(val: string, offset: int = 1) {
    return isWrappedWith(val, '"', offset) || isWrappedWith(val, "'", offset);
}
````

I'm sure you needed something like this in many places.

File: src/utils/angular.ts [hunk 8]
```diff
@@ -81,11 +83,9 @@ export function isAngularDirective(name: string): boolean {
...
 	return (
 		val.length >= 5 &&
-		((val[0] === '"' && val[val.length - 1] === '"') || (val[0] === "'" && val[val.length - 1] === "'")) &&
-		val[1] === '{' &&
-		val[2] === '{' &&
-		val[val.length - 2] === '}' &&
-		val[val.length - 3] === '}' &&
+		isQuoted(val) &&
+		isWrappedWith(val, '{', '}', 1) &&
+		isWrappedWith(val, '{', '}', 2) &&
```
ST-DDT, line 88:
```suggestion
		isWrappedWith(val, '{{', '}}', 1) &&
```

ST-DDT (changes requested):

src/utils/svelte.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * In this case `val` is `"{ cat.id }"`.
+ *
+ * ---
+ *
+ * @param val Value of tag attribute.
+ * @returns `true` if `val` passes the svelte interpolation check, otherwise `false`.
+ */
+export function isSvelteInterpolation(val: string): boolean {
+	return val.length >= 3 && isQuoted(val) && isWrappedWith(val, '{', '}', 1) && !val.includes('{', 2);
```
ST-DDT, line 21: The expression `isWrappedWith(val, '{', '}', 1) && !val.includes('{', 2)` seems to be repeated similarly in the angular.ts.

How about extracting it into in another method?

````js
function isWrappedWithExclusive(val: string, start: string, end: string, offset: number) {
    return isWrappedWith(val, start, end, offset) && !val.includes(start, start.length + offset);
}
````

ST-DDT (changes requested):

src/utils/svelte.ts
Shinigami92: Theoretically a good idea, but IMO we start to over-engineer this :smile:  
Let's let it as is for now and if there are more cases in the future, I will rethink about this extraction
But until now, it's already a one-liner :sweat_smile:

src/printer.ts
Shinigami92: Okay, this code hurts, but I personally decided for now to await feedback from svelte community if there are further errors and printed warnings in their console output
Based on this feedback I can refactor this messy code
But now it's working and the svelte community can start working our formatter

PR #208, opened 2021-05-01
Support bracket spacing for interpolated-code token

Shinigami92:


src/utils/common.ts
```diff
@@ -152,10 +152,13 @@ export function isMultilineInterpolation(val: string): boolean {
...
  * @param code Code that is enclosed in brackets.
+ * @param param2 Brackets.
+ * @param param2."0" Opening brackets.
+ * @param param2."1" Closing brackets.
  * @returns The handled string.
  */
-export function handleBracketSpacing(bracketSpacing: boolean, code: string): string {
-	return bracketSpacing ? `{{ ${code} }}` : `{{${code}}}`;
+export function handleBracketSpacing(bracketSpacing: boolean, code: string, [opening, closing] = ['{{', '}}']): string {
+	return bracketSpacing ? `${opening} ${code} ${closing}` : `${opening}${code}${closing}`;
```
Shinigami92, line 161: Maybe we should use another kind of parameter like `1` or `2` :thinking:  
But this could result in more processing
Maybe a benchmark could help here

Issue #241, opened 2021-06-27
feature request: explicit div in output


bitencode:
## Request / Idea

Option to not remove the `div` element when literal class or id is used.
I recently found your plugin and it's just about perfect for our needs (Vue with Pug).  However, for reasons we don't allow the use of `.my-class` or `#myId` in our templates.  We require the use of `div.my-class` particularly because it is hard for new devs to grok and it's more explicit.

## Input

```pug
div.my-class(attrs)
  div(class="anohterClass")
```

## Expected Output

```pug
div.my-class(attrs)
  div.anotherClass
```
instead this is always produced:
```pug
.my-class(attrs)
  .anotherClass
```

While that is perfectly valid pug syntax we don't allow it because we believe it's not a good feature.

## Additional Context
@Shinigami92 Would you be open to  a PR that added a pug option to allow the `div` element to remain in the output?
If so, do you have an option name preference? I was thinking of `pugExplicitDiv`, but will use your preference.

I started looking at this and it seems like it would be fairly easy to create.

Shinigami92:
This will land in the next release, but I have some other PRs that also needs to be merged
Maybe I will release a beta in the next few days

bitencode:
Thank you @Shinigami92

Shinigami92:
Please try out the new beta [v1.16.0-0](https://www.npmjs.com/package/@prettier/plugin-pug/v/1.16.0-0) and report anything back :slightly_smiling_face:

bitencode:
Hi @Shinigami92, now I'm on vacation this week, but have a laptop with me 😉.  I haven't found any issues related to the `pugExplicitDiv` feature, other than the release notes say `pubExplicitDiv` instead of `pugExplicitDiv`.

I ran 1.16.0-0 on two of my personal projects and found no issues it all, but both projects have had a single developer and are pretty consistent in style.  One has about 50 components and the other about 75 and `pugExplicitDiv` will correctly add or remove the `div` based on the setting.

I also ran 1.16.0-0 on our SDVI front-end code base and found no issues with `pugExplicitDiv` (worked as expected, found some `div`s that were missed and automatically added them.  I discarded changes and went the other way and it correctly removed all the `div` that had `id` or `class`, so I think we're good 👍 

[...]

PR #242, opened 2021-06-27
feat: explicit div pug option

bitencode:
# Changes:
- added a `pugExplicitDiv` option (defaults to `false`, current behavior)
- added tests to show option changes the output
- added an entry in pug specific docs

### Option `pugExplicitDiv: false` (or no entry)

Input:
```pug
div.myClass#myId
```
Output:
```pug
.myClass#myId
```

#Option `pubExplicitDiv: true`

Input:
```pug
div.myClass#myId
```
Output:
```pug
div.myClass#myId
```

# Additional Information:
This turned out to be easier than I thought, so decided to go ahead and do a PR in hopes this is an acceptable option.
I looked around at other options and the code and tried to match what it currently here.  Let me know if I missed something or you'd like changes.
The `yarn.lock` file changed because I ran the `./release.sh` script as per 'How to contribute".  I can revert that if it wasn't supposed to change.
I tried to follow the current pattern/language for the help doc guide entry.

Fulfills feature request at #241

Thank you!
Loving this plug-in! ❤️ 

PS. I am unable to add you as a reviewer so just mentioning you instead - thanks @Shinigami92


Shinigami92:
I'm currently only on mobile, cause I'm in vacation 
Maybe I can make an in depth review tomorrow 

Could you move the test actual and expected to pug files? And then add a case for testing wrapping? So e.g. a line wraps it's attributes after 80 chars and the line with div has 3 more chars so it has e.g. 81 instead of 78 chars

docs/guide/pug-specific-options.md
```diff
@@ -465,3 +465,38 @@ Disables id formatting.
...
+
+Define if a `div` element should *always* be printed with literal class and id formatting.
+
+### Options
+
+**Type:** `boolean`  
+**Default:** `false`
+
+#### `false`
+Input:
```
Shinigami92, line 485: I think you can run `yarn format` to apply format to all files

src/options/pug-explicit-div.ts
```diff
@@ -0,0 +1,11 @@
+import type { BooleanSupportOption } from 'prettier';
+import { CATEGORY_PUG } from '.';
+
+/** Pug default div element. */
+export const PUG_EXPLICIT_DIV: BooleanSupportOption = {
+	since: '1.16.0',
+	category: CATEGORY_PUG,
+	type: 'boolean',
+	default: false,
+	description: 'Include `div` element when followed by literal class or id syntax'
```
Shinigami92, line 10:
```suggestion
	description: 'Include `div` tag when followed by literal class or id syntax'
```

Shinigami92 (changes requested):

bitencode:
hi @Shinigami92, thanks for the testing feedback - excellent idea as I found some issues 😳.
My original commit only addressed not *removing* `div` from the markup and I hadn't considered the need to *add* `div`s to the output if they were not in the input.  That was a little more invasive and took a lot more digging around and testing to find a solution.
All the test changes are in commit [99e548c](https://github.com/prettier/plugin-pug/pull/242/commits/99e548c43298f3507dd06df1e8782993392f3e37) and all code changes in `printer.ts` are in  [d1af1f5](https://github.com/prettier/plugin-pug/pull/242/commits/d1af1f53f6f38038e40deb83813b93e38931212f).
thanks!

docs/guide/pug-specific-options.md
```diff
@@ -465,3 +465,46 @@ Disables id formatting.
...
+```pug
+.foo.bar
+```
+
+#### `true`
+
+Input:
+
+```pug
+div.foo.bar
```
Shinigami92, line 503:
```suggestion
.foo.bar
```

To make it even more clear, we should remove the tag here as it will be added if it doesn't exist

Otherwise we would need something like always, as-is and never
But I think this will never be requested by someone, and IMO it doesn't make sense to provide an option like as-is

src/options/pug-explicit-div.ts
```diff
@@ -0,0 +1,11 @@
+import type { BooleanSupportOption } from 'prettier';
+import { CATEGORY_PUG } from '.';
+
+/** Pug default div element. */
```
Shinigami92, line 4:
```suggestion
/** Pug default div tag. */
```

src/printer.ts
```diff
@@ -1012,8 +1020,10 @@ export class PugPrinter {
 				case 'newline':
 				case 'outdent':
 				case 'indent': {
-					this.possibleIdPosition = this.result.length + this.computedIndent.length;
-					const result: string = `${this.computedIndent}${val}`;
+					const optDiv: string = this.options.pugExplicitDiv ? 'div' : '';
```
Shinigami92, line 1023: Could you write this out to optionalDiv? (Occurs in some other lines too)

tests/options/pugExplicitDiv/pug-explicit-div.test.ts
```diff
@@ -0,0 +1,48 @@
...
+			});
+
+			expect(actual).toBe(expected);
+		});
+
+		test('should handle pugExplicitDiv:false', () => {
+			const expected: string = readFileSync(resolve(__dirname, 'formatted-implicit-div.pug'), 'utf8');
+			const actual: string = format(code, {
+				// @ts-expect-error
+				pugExplicitDiv: false,
```
Shinigami92, line 30: Explicit "override" values should be placed after spreading so it will always be taken, also if someone write this value in the common object

tests/options/pugExplicitDiv/unformatted.pug
```diff
@@ -0,0 +1,12 @@
+div.class(id="id" attribute="value") div with id attr test
```
Shinigami92, line 1: Maybe you can get rid of this file and rename the other two
Then you can use the always against the never case and the never case against the always 🤔

Only thing that wouldn't be tested then would be mixed -> formatted result 🤔

---

I think I have some similar tests, could you search a little bit and adept to a similar solution?
It's a little bit hard for me, cause I'm just on my phone right now 😅
Will be fully back from vacation late evening on next friday

Shinigami92 (changes requested):

docs/guide/pug-specific-options.md
bitencode: Agreed 😄 - the whole point of Prettier is to make code consistent so, always on or always off makes sense - a mix seems like the wrong direction. 👍

tests/options/pugExplicitDiv/pug-explicit-div.test.ts
bitencode: oops, yes - will fix.

tests/options/pugExplicitDiv/unformatted.pug
bitencode: Okay, I'll look around and see what else I find that you have.  I was mostly following what I found in `pugClassNotation` that had:
* unformatted.pug
* formatted-attribute.pug
* formatted-literal.pug

I did it because I found in our code base a mixed set (someone managed to commit with no leading `div` and no one noticed).  I can see how the attribute and literal could be used to test each other though.  For development of the tests it was helpful to have the source and two targets so I could find where problems where.

You shouldn't review code while on vacation 🌴 - you should relax.  This can wait until you're back.

src/printer.ts
```diff
@@ -721,7 +729,7 @@ export class PugPrinter {
 			logger.debug('after token', this.currentLineLength);
 			if (hasLiteralAttributes) {
 				// Remove div as it will be replaced with the literal for id and/or class
-				if (this.previousToken?.type === 'tag' && this.previousToken.val === 'div') {
+				if (this.previousToken?.type === 'tag' && this.previousToken.val === 'div' && !this.options.pugExplicitDiv) {
```
Shinigami92, line 732:
```suggestion
				if (
				  this.previousToken?.type === 'tag' &&
				  this.previousToken.val === 'div' &&
				  !this.options.pugExplicitDiv
				) {
```

tests/options/pugExplicitDiv/pug-explicit-div.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+describe('Options', () => {
+	describe('pugExplicitDiv', () => {
+		const code: string = readFileSync(resolve(__dirname, 'unformatted.pug'), 'utf8');
+		const commonOpts = {
+			parser: 'pug',
+			plugins: [plugin],
+			// use this common options in all tests to force specific wrapping
+			pugAttributeSeparator: 'none',
+			pugPrintWidth: 80,
```
Shinigami92, line 14:
```suggestion
			pugPrintWidth: 80
```

tests/options/pugExplicitDiv/pug-explicit-div.test.ts
```diff
@@ -0,0 +1,48 @@
...
+		}
```
Shinigami92, line 15:
```suggestion
		};
```

tests/options/pugExplicitDiv/pug-explicit-div.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+		test('should handle unspecified pugExplicitDiv', () => {
+			const expected: string = readFileSync(resolve(__dirname, 'formatted-implicit-div.pug'), 'utf8');
+			const actual: string = format(code, {
+				...commonOpts,
```
Shinigami92, line 20:
```suggestion
				...commonOpts
```

tests/options/pugExplicitDiv/pug-explicit-div.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+			expect(actual).toBe(expected);
+		});
+
+		test('should handle pugExplicitDiv:false', () => {
+			const expected: string = readFileSync(resolve(__dirname, 'formatted-implicit-div.pug'), 'utf8');
+			const actual: string = format(code, {
+				...commonOpts,
+				// @ts-expect-error
+				pugExplicitDiv: false,
```
Shinigami92, line 31:
```suggestion
				pugExplicitDiv: false
```

tests/options/pugExplicitDiv/pug-explicit-div.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+			expect(actual).toBe(expected);
+		});
+
+		test('should handle pugExplicitDiv:true', () => {
+			const expected: string = readFileSync(resolve(__dirname, 'formatted-explicit-div.pug'), 'utf8');
+			const actual: string = format(code, {
+				...commonOpts,
+				// @ts-expect-error
+				pugExplicitDiv: true,
```
Shinigami92, line 42:
```suggestion
				pugExplicitDiv: true
```

src/printer.ts
```diff
@@ -721,7 +731,11 @@ export class PugPrinter {
 			logger.debug('after token', this.currentLineLength);
 			if (hasLiteralAttributes) {
 				// Remove div as it will be replaced with the literal for id and/or class
-				if (this.previousToken?.type === 'tag' && this.previousToken.val === 'div') {
+				if (
+				  this.previousToken?.type === 'tag' &&
+				  this.previousToken.val === 'div' &&
+				  !this.options.pugExplicitDiv
```
Shinigami92, line 737:
```suggestion
					!this.options.pugExplicitDiv
```

src/printer.ts
```diff
@@ -721,7 +731,11 @@ export class PugPrinter {
 			logger.debug('after token', this.currentLineLength);
 			if (hasLiteralAttributes) {
 				// Remove div as it will be replaced with the literal for id and/or class
-				if (this.previousToken?.type === 'tag' && this.previousToken.val === 'div') {
+				if (
+				  this.previousToken?.type === 'tag' &&
+				  this.previousToken.val === 'div' &&
```
Shinigami92, line 736:
```suggestion
					this.previousToken.val === 'div' &&
```

src/printer.ts
```diff
@@ -721,7 +731,11 @@ export class PugPrinter {
 			logger.debug('after token', this.currentLineLength);
 			if (hasLiteralAttributes) {
 				// Remove div as it will be replaced with the literal for id and/or class
-				if (this.previousToken?.type === 'tag' && this.previousToken.val === 'div') {
+				if (
+				  this.previousToken?.type === 'tag' &&
```
Shinigami92, line 735:
```suggestion
					this.previousToken?.type === 'tag' &&
```

src/printer.ts
```diff
@@ -377,6 +378,10 @@ export class PugPrinter {
 	}
 
 	private replaceTagWithLiteralIfPossible(search: RegExp, replace: string): void {
+		if (this.options.pugExplicitDiv) {
+		  return;
```
Shinigami92, line 382:
```suggestion
			return;
```

tests/options/pugExplicitDiv/pug-explicit-div.test.ts
```diff
@@ -0,0 +1,48 @@
+import { readFileSync } from 'fs';
+import { resolve } from 'path';
+import { format } from 'prettier';
+import { plugin } from '../../../src/index';
+
+describe('Options', () => {
+	describe('pugExplicitDiv', () => {
+		const code: string = readFileSync(resolve(__dirname, 'unformatted.pug'), 'utf8');
+		const commonOpts = {
```
Shinigami92, line 9: I think you need to find a suitable typedef here. Maybe it's the second parameter of prettiers `format` function.

bitencode: Sorry for the tab and typescript lint issue.  I forgot to run `yarn lint` - in all the repos I work in it is automatic before commit and `prettier` catches most our stuff while still in the editor.  First type writing typescript as my company does not use it.

Shinigami92 (approved):

PR #249, opened 2021-07-03
Handle non-JS script content

OiYouYeahYou:
@Shinigami92 I ended up doing most this before you PR'd. Take what you like from it or close it.

To add the I collated all the current IANA types in a single file: [media-types.txt](https://github.com/prettier/plugin-pug/files/6759064/media-types.txt) from the official [IANA reference](https://www.iana.org/assignments/media-types/media-types.xhtml). But most of them are not textual so have nothing to do with `<script>` tags. I've tried to find some sort of reference collating what people use the `type` attribute. Apart from refs to the old `language` attribute and VBScript, JScript and TCL and these aren't worth supporting.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -94,6 +94,7 @@ module.exports = defineConfig({
 					'coffeescript',
 					'colno',
 					'doctype',
+					'ecmascript',
 					'endregion',
 					'eos',
 					'fallthrough',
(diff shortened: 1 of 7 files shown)


Shinigami92:
Thanks, I think tomorrow I will take some from both PRs and so I have the best as result

Shinigami92 (approved):
I reviewed you code now in depth and wow, good work!
I will take this PR as is and extend a bit based on it :slightly_smiling_face:

Issue #250, opened 2021-07-05
Bug: corrupts pug variable names that are assigned to bound Vue props


bitencode:
### Plugin Version

v1.15.3 & 1.16.0-0

### Prettier Version

v2.3.1

### Which frameworks are affected?

- [ ] none
- [X] vue
- [ ] angular
- [ ] svelte

### Node Version

v12.19.1

### Which operating systems have you used?

- [ ] Linux
- [X] macOS
- [ ] Windows

### Prettier config

```json
{
  arrowParens: 'always',
  bracketSpacing: true,
  jsxBracketSameLine: false,
  printWidth: 100,
  proseWrap: 'preserve',
  semi: true,
  singleQuote: true,
  tabWidth: 2,
  trailingComma: 'es5',
  useTabs: false,
  pugAttributeSeparator: 'none',
  pugClassNotation: 'literal',
  pugEmptyAttributes: 'none',
  pugExplicitDiv: true,
  pugFramework: 'vue',
  pugIdNotation: 'literal',
  pugSingleFileComponentIndentation: true,
  pugSingleQuote: false,
  pugSortAttributes: 'asc',
  pugSortAttributesBeginning: ['^:?is', '^v-for', '^v-', '^:?id$', '^:?key$', '^:?ref$'],
  pugSortAttributesEnd: ['^:', '^v-on$', '^@', '^v-html$', '^v-text$'],
  pugWrapAttributesThreshold: 1,
}
```

### Input

```pug
- const circleSize = 72
- const circleWidth = 7
- v-progress-circular(
  :color="getProgressColor(currentCapacity)"
  :size=circleSize
  :value="capacityPercentage"
  :width=circleWidth
) {{ currentCapacity }}/{{ maxCapacity }}
```

### Output or Error

```pug
- const circleSize = 72
- const circleWidth = 7
v-progress-circular(
  :color="getProgressColor(currentCapacity)"
  :size="ircleSiz"
  :value="capacityPercentage"
  :width="ircleWidt"
) {{ currentCapacity }}/{{ maxCapacity }}
```

### Expected Output

```pug
- const circleSize = 72
- const circleWidth = 7
v-progress-circular(
  :color="getProgressColor(currentCapacity)"
  :size=circleSize
  :value="capacityPercentage"
  :width=circleWidth
) {{ currentCapacity }}/{{ maxCapacity }}
```

### Additional Context

[...]

PR #252, opened 2021-07-06
Fix corrupt attribute values on dynamically bound Vue props without quotes

bitencode:
# Changes:
- First commit adds a unit test to expose issue #250.
- Second commit changes code so that new test passes and all other test still pass (checks for quotes before stripping them).

Note that I changed `isQuoted` and added the back-tic quote to the test with the other two quote styles (was `"'`, now is ``"`'``). All the tests pass, running this on my code based produced correct results, but I only have plain and Vue-based code (no angular, react, svelte, etc..).  I looked at all the calling sites of `isQuoted` and it looks okay to me, but someone more familiar with the code base should probably review that carefully.

### DIFF
--- a/src/utils/common.ts
+++ b/src/utils/common.ts
@@ -165,7 +165,7 @@ export function isWrappedWith(val: string, start: string, end: string, offset: n
  * @returns Whether the value is quoted or not.
  */
 export function isQuoted(val: string): boolean {
-	if (/^(["'])(.*)\1$/.test(val)) {
+	if (/^(["'`])(.*)\1$/.test(val)) {
 		// Regex for checking if there are any unescaped quotations.
 		const regex: RegExp = new RegExp(`${val[0]}(?<!\\\\${val[0]})`);
 		return !regex.test(val.slice(1, -1));
--- a/tests/issues/issue-250/formatted.pug
+++ b/tests/issues/issue-250/formatted.pug
@@ -1,7 +1,9 @@
+- const circleSize = 72;
+- const circleWidth = 7;
 div
   v-progress-circular(
     :color="getProgressColor(currentCapacity)",
-    :size="circleSize",
+    :size=circleSize,
     :value="capacityPercentage",
-    :width="circleWidth"
+    :width=circleWidth
   ) {{ currentCapacity }}/{{ maxCapacity }}
--- a/tests/issues/issue-250/formatted.pug
+++ b/tests/issues/issue-250/formatted.pug
@@ -1,0 +1,7 @@
+div
+  v-progress-circular(
+    :color="getProgressColor(currentCapacity)",
+    :size="circleSize",
+    :value="capacityPercentage",
+    :width="circleWidth"
+  ) {{ currentCapacity }}/{{ maxCapacity }}
--- a/tests/issues/issue-250/issue-250.test.ts
+++ b/tests/issues/issue-250/issue-250.test.ts
@@ -4,7 +4,7 @@ import { format } from 'prettier';
 import { plugin } from './../../../src/index';
 
 describe('Issues', () => {
-	test('should handle spread destructuring in Vue syntax', () => {
+	test('should not quote pug variables assigned to Vue props', () => {
 		const expected: string = readFileSync(resolve(__dirname, 'formatted.pug'), 'utf8');
 		const code: string = readFileSync(resolve(__dirname, 'unformatted.pug'), 'utf8');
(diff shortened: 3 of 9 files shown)


src/utils/common.ts
```diff
@@ -165,7 +165,7 @@ export function isWrappedWith(val: string, start: string, end: string, offset: n
  * @returns Whether the value is quoted or not.
  */
 export function isQuoted(val: string): boolean {
-	if (/^(["'])(.*)\1$/.test(val)) {
+	if (/^(["'`])(.*)\1$/.test(val)) {
```
Shinigami92, line 168: If I remember correctly, backtick quoted strings are pre processed by pug as JS. Therefore it is handled a bit differently than normal quoted strings.
But beside that, if this is the case, I should find a test case to cover that 🤔 
Will have a deeper look later.

tests/issues/issue-250/formatted.pug
```diff
@@ -0,0 +1,7 @@
+div
+  v-progress-circular(
+    :color="getProgressColor(currentCapacity)",
+    :size="circleSize",
```
Shinigami92, line 4: I need to look into this later cause your assumption could be wrong
`:size=circleSize` should be converted to `:size=circleSize` and not `:size="circleSize"`, cause `circleSize` could be a variable in the pug file itself 🤔

tests/issues/issue-250/issue-250.test.ts
```diff
@@ -0,0 +1,14 @@
+import { readFileSync } from 'fs';
+import { resolve } from 'path';
+import { format } from 'prettier';
+import { plugin } from './../../../src/index';
+
+describe('Issues', () => {
+	test('should handle spread destructuring in Vue syntax', () => {
```
Shinigami92, line 7: Looks like a copy paste issue to me 😛

bitencode: Um, yes 😳

tests/issues/issue-250/formatted.pug
bitencode: You are so right... that's what I get for reviewing code on a laptop while on vacation 😳.  Yes, what I did here was the wrong fix, they shouldn't have quotes, they should be left alone (still a corruption problem, but results need to be different).  I scanned up further in our source file and found this:
```pug
<template lang="pug">
  - const circleSize = 72;
  - const circleWidth = 7;
```
which is very atypical in our Vue code, but still makes sense here.

src/utils/common.ts
Shinigami92: @bitencode Do you have a case for this? Then add it to the tests
If not, I would like to revert this line and just change it when it is absolutely necessary

bitencode: @Shinigami92 I do not have a specific test case for this. However, with the changes to avoid quoting pug variables assigned to Vue props to other test cases break (`issue-61` - 'should not wrap chained data-bindings' and `issue-80` - 'should not reformat multiline interpolation strings' because they are then not unquoted before being passed to the prettier `format` function.

It's possible that if `isQuoted` need to *not* check for back-tic quotes that there should be another function that does the check needed in `formatDelegatePrettier`.  Or possibly an additional parameter that indicates which quotes to check for with the default being `"'` and the new call can pass in all 3 types of quotes.  I don't know this code base well, or your coding preferences, so I'm open to suggestions.

Shinigami92: Thx :slightly_smiling_face: I will look into the branch and discover it
Will send feedback later back to you

Shinigami92: AHA! The thing is, that 61 and 80 have backtick quoted strings, but can be normal strings cause they don't use code interpolation! So no `${` in it.
Then my plugin converts these template literal strings to normal strings (by design)

But I currently don't understand why the new change is needed then :thinking: 

And we should create a test to check that it doesn't convert the backticks if it contains interpolated code in one single line

Shinigami92: @bitencode So I added commit: https://github.com/prettier/plugin-pug/pull/252/commits/a491e08c6596e2871ced6c5e78e9b61bcfa261d5

Seems that it still works with your change and not work when reverting...
So thx I'm now less afraid and confident using your change :slightly_smiling_face:

Shinigami92 (approved):

src/utils/common.ts
bitencode: 👍 Awesome - Thanks!

PR #256, opened 2021-07-10
Prevent wrong quotation

### LINKED ISSUE #255: Bug: pugAttributeSeparator='as-needed' breaks quotes of the preceding attribute value
### Plugin Version

v1.16.0

### Prettier Version

v2.3.2

- [X] none
- [ ] vue
- [ ] angular
- [ ] svelte

### Node Version

v10.19.0

- [X] Linux
- [ ] macOS
- [ ] Windows

### Prettier config

```json
overrides:
  - files: "*.pug"
    options:
      semi: false
      singleQuote: true
      trailingComma: none
      arrowParens: avoid
      pugAttributeSeparator: as-needed
```

### Input

```pug
doctype html
html
  head
    title= title
    link(rel='stylesheet' , href='/stylesheets/style.css')
  body
    block content
```

### Output or Error

```bash
doctype html
html
  head
    title= title
    link(rel="stylesheet" href='/stylesheets/style.css')
  body
    block content
```

### Expected Output

```pug
doctype html
html
  head
    title= title
    link(rel='stylesheet' href='/stylesheets/style.css')
  body
    block content
```

### Additional Context

When using the option _pugAttributeSeparator: as-needed_, the pugin breaks quotes around attribute values, which are followed by atleast one space before comma and the following attribute key/value pair.
This can seen in the 5th line of my example:

> link(rel='stylesheet' , href='/stylesheets/style.css')

will get rewritten by the plugin as:

> link(rel="stylesheet" href='/stylesheets/style.css')

When I rerun it the 2nd time, it ends up correctly, as:

> link(rel='stylesheet' href='/stylesheets/style.css')

With more than 2 attributes, it only breaks the ones which have atleast one space before the comma and the following attributes, but correctly sets up the attribues, which are immediately followed by a comma.

Without the _singleQuote: true_ option, it will break the quotes to the single quotes. Put simply, it always breaks the quotes since it applies exactly the opposite of what you have set.

Shinigami92:
closes #255

### DIFF
--- a/src/printer.ts
+++ b/src/printer.ts
@@ -916,30 +916,34 @@ export class PugPrinter {
 				val = this.formatSvelteInterpolation(val);
 			} else if (isStyleAttribute(token.name, token.val)) {
 				val = this.formatStyleAttribute(val);
-			} else if (isQuoted(val)) {
-				val = makeString(val.slice(1, -1), this.quotes);
-			} else if (val === 'true') {
-				// The value is exactly true and is not quoted
-				return;
-			} else if (token.mustEscape) {
-				val = format(val, { parser: '__js_expression', ...this.codeInterpolationOptions });
-
-				const lines: string[] = val.split('\n');
-				const codeIndentLevel: number = this.wrapAttributes ? this.indentLevel + 1 : this.indentLevel;
-				if (lines.length > 1) {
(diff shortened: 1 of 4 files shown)


ST-DDT (approved):
Looks good to me

PR #270, opened 2021-08-03
Use pugTabWidth for calculating line length

### LINKED ISSUE #268: Bug: Incorrect wrapping when `useTabs: true`, assumes tab is 0 width
## Info

| Tool      | Version   |
| --------- | --------- |
| Plugin    | ^1.16.0-0 |
| Prettier  | ^2.3.2    |
| Framework | none      |
| Node      | 12        |
| OS        | linux     |

## Prettier config

```yaml
useTabs: true
tabWidth: 8 # Just for example
```

## Input

```pug
p
	input#some-very-long-id-for-example-sake(type="input", placeeholder="example")
```

## Output or Error

```pug
p
	input#some-very-long-id-for-example-sake(type="input", placeeholder="example")
```

Formatting with spaces instead of tabs
```pug
p
        input#some-very-long-id-for-example-sake(
                type="input",
                placeeholder="example"
        )
```

## Expected Output

```pug
p
	input#some-very-long-id-for-example-sake(
		type="input",
		placeeholder="example"
	)
```

## Additional Context

When using `useTabs: false`, it wraps correctly. Otherwsie wrapping doesn't account of the indent width and just assumes 0-width tabs.

Shinigami92:
closes #268

### DIFF
--- a/tests/issues/issue-268/formatted.pug
+++ b/tests/issues/issue-268/formatted.pug
@@ -21,9 +21,19 @@ html
 				form(name="contact", method="POST")
 					.row.gtr-uniform
 						.col-6.col-12-x-small
-							input#name(type="text", name="name", placeholder="Name *", required)
+							input#name(
+								type="text",
+								name="name",
+								placeholder="Name *",
+								required
+							)
 						.col-6.col-12-x-small
-							input#email(type="email", name="email", placeholder="Email *", required)
+							input#email(
+								type="email",
+								name="email",
+								placeholder="Email *",
+								required
+							)
 						.col-12
 							input#subject(
 								type="text",
--- a/tests/issues/issue-268/formatted.pug
+++ b/tests/issues/issue-268/formatted.pug
@@ -1,5 +1,46 @@
-p
-	input#some-very-long-long-id-for-example-sake(
-		type="input",
-		placeholder="example"
-	)
+doctype html
+html
+	head
+	body
+		#indentation-tests
+			.short(style="display: none") Hello World
+			.getting-longer(style="display: none") I can't just say "Hello World" here
+			.i-am-not-being-creative-with-these-names(
+				style="display: none",
+				color="red"
+			).
+				There is surely more I could write here
+			.short-again(style="display: none", color="red", font="Sans Serif")
+				| I could surely more I could write here
+
+			.hi(data-wrap="false") This should be wrapped
+
+		section#contact
+			.container
+				h3 Contact
+				form(name="contact", method="POST")
+					.row.gtr-uniform
+						.col-6.col-12-x-small
+							input#name(type="text", name="name", placeholder="Name *", required)
+						.col-6.col-12-x-small
+							input#email(type="email", name="email", placeholder="Email *", required)
+						.col-12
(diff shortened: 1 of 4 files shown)


Shinigami92:
@OiYouYeahYou Could you check what I need to adjust in the test to make it failing?
Cause currently everything is green :thinking:

tests/issues/issue-268/formatted.pug
```diff
@@ -0,0 +1,46 @@
...
+
+		section#contact
+			.container
+				h3 Contact
+				form(name="contact", method="POST")
+					.row.gtr-uniform
+						.col-6.col-12-x-small
+							input#name(type="text", name="name", placeholder="Name *", required)
+						.col-6.col-12-x-small
+							input#email(type="email", name="email", placeholder="Email *", required)
```
Shinigami92, line 26: @OiYouYeahYou do you mean these two lines?

image

I see that they are longer than 80 cols :thinking:  
And you want it that they wrap?

OiYouYeahYou: It differs depending on what indentation width you are using, but yes those two lines are a problem. In my opinion the formatting should wrap identically to the spaces.

The test I've been using replaces tabs for spaces at the same with and then compares them:

```ts
function comparisonTabTest(message: string, { output, tabWidth }: { output: string; tabWidth: number }): void {
	test(message, () => {
		const spaces: string = format(unformatted, { parser: 'pug', plugins: [plugin], tabWidth, useTabs: false });
		const tabs: string = format(unformatted, {
			parser: 'pug',
			plugins: [plugin],
			tabWidth,
			useTabs: true
		}).replace(/\t/g, ' '.repeat(tabWidth));

		expect(tabs).toBe(spaces);
	});
}
```

I've already looked at fixing this and it's non-trivial to repair at the moment. There is a hacky solution though, which is to use `useTabs: false` and then convert the spaces to tabs after printing.

Shinigami92: Okay now I fully understood the original issue ^^

Yeah, we have a `currentLineLength`

https://github.com/prettier/plugin-pug/blob/998a192e7d1f5ba1af4f98aacf0448dbcc26364a/src/printer.ts#L174

And then we have lines like 

https://github.com/prettier/plugin-pug/blob/998a192e7d1f5ba1af4f98aacf0448dbcc26364a/src/printer.ts#L1000

where we using `indentString.length`

But it only counts as `1` instead of e.g. `8` for each `\t`.
So we could try to create a `get indentStringLength() {...}` and use it instead.

You have write access to this branch `issue-268`. Would you like to try it out?
I don't have the time to code for the next ~40 hours :slightly_frowning_face:

Shinigami92: That was much more easy than I thought :eyes:

bminer:
Something is still not right. Also [lines 7 and 13 in the test](https://github.com/prettier/plugin-pug/pull/270/files#diff-5c83253b44fcd0ba64f8ae36ca928a7ff97b372b3d62d622d8df9e30982b53a2R7-R13) should wrap at tab width 8, for example.

PR #290, opened 2021-09-08
Handle Vue slot shorthand syntax

### LINKED ISSUE #289: Bug: Can not handle Vue slot shortcut syntax
### Plugin Version

master

### Prettier Version

^2.3

- [ ] none
- [X] vue
- [ ] angular
- [ ] svelte

### Node Version

14

- [ ] Linux
- [ ] macOS
- [X] Windows

### Prettier config

```json
{
  printWidth: 120,
  singleQuote: true,
  trailingComma: 'all',
  pugAttributeSeparator: 'none',
  pugFramework: 'vue',
  pugSingleFileComponentIndentation: true,
  pugSingleQuote: false,
}
```

### Input

```pug
my-component(foo="bar", #default="{ baz }")
```

### Output or Error

```bash
my-component(foo="bar", #default="{ baz }")
```

### Expected Output

```pug
my-component(foo="bar" #default="{ baz }")
```

### Additional Context

it just does not format the content at all (via VS code). removing the destructuring will make it work

Shinigami92:
close #289

### DIFF
--- a/tests/issues/issue-289/unformatted.pug
+++ b/tests/issues/issue-289/unformatted.pug
@@ -1,0 +1,1 @@
+my-component( foo="bar",  #default="{  baz }" )
--- a/tests/issues/issue-289/unformatted.pug
+++ b/tests/issues/issue-289/unformatted.pug
@@ -1,1 +1,1 @@
-my-component( foo="bar",  #default="{  baz }" )
+my-component( #default="{  foo,  bar:  baz }")
--- a/src/utils/vue.ts
+++ b/src/utils/vue.ts
@@ -50,7 +50,7 @@ export function isVueEventBinding(name: string): boolean {
  * @returns `true` if `name` passes the vue expression check, otherwise `false`.
  */
 export function isVueExpression(name: string): boolean {
-	return /^((v-(bind|slot))?:|v-(model|slot|if|for|else-if|text|html)).*/.test(name);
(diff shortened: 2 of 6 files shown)


tests/issues/issue-289/unformatted.pug
```diff
@@ -0,0 +1 @@
+my-component( foo="bar",  #default="{  baz }" )
```
Shinigami92, line 1: @backbone87 Please provide an example that fails the test 🤔

Shinigami92: If you're using Vue 3, could you try to remove/disable `pugSingleFileComponentIndentation`? Just guessing that maybe there is a problem with that?! :thinking: But not sure.

backbone87: failing:

unformatted: `my-component(  #default="{ foo, bar: baz }")`
formatted: `my-component(#default="{ foo, bar: baz }")`

```
SyntaxError: Unexpected token ','

      301 |                             }
      302 |                     } catch (error: any) {
    > 303 |                             throw new Error(error);
          |                                   ^
      304 |                     }
      305 |                     token = this.getNextToken();
      306 |             }
```

backbone87: the test passes with the following change:
```ts
export function isVueExpression(name: string): boolean {
	return /^((v-(bind|slot))?:|v-(model|slot|if|for|else-if|text|html)|#).*/.test(name);
}
```
(added `|#`)

Issue #45, opened 2019-11-15
Bug: Doesn't take the singleQuote option


Applelo:
Since the 1.0.10, singlequote are transformed into double quotes even if I set singleQuote option.
I tried with the 1.0.9 and I don't have this bug.

For example:
```
- const faker = require('faker')
```

is transformed into
```
- const faker = require("faker")
```

Shinigami92:
Is it only in these code blocks?

Applelo:
No, in all my pug files, single quotes are tranformed into double quotes.

Shinigami92:
Do you have a repo that you can link?
And can you copy your prettier config?

Shinigami92:
I can't reproduce it? I have a project with 81 pug files

My `.prettierrc.json` config
```json
{
  "arrowParens": "always",
  "bracketSpacing": true,
  "printWidth": 120,
  "semi": true,
  "singleQuote": true,
  "trailingComma": "none",
  "tabWidth": 2,
  "useTabs": false,
  "overrides": [
    {
      "files": ["*.pug"],
      "options": {
        "tabWidth": 2,
        "singleQuote": false
      }
    }
  ]
}
```

Shinigami92:
I also have tests for this: https://github.com/prettier/plugin-pug/tree/master/test/quotes

Applelo:
I have create a repo with an example of the problem. The prettier config is in the package.json.
https://github.com/Applelo/plugin-pug-issue-45

Shinigami92:
Ha! I got it

The code in your pug file is JavaScript 
JavaScript will be single quoted 
And pug attributes will be double quoted

Just add a tag with attributes and play with the singleQuote attribute in both places 
Then you will understand

Applelo:
Oh ok, so I add this configuration and it's work fine. Thanks for your answer.
```
{
  "files": "*.js",
  "singleQuote": true
},
```

Applelo:
For me, pug attributes are singleQuoted and javascript content is transformed into double quotes.

Shinigami92:
https://github.com/prettier/plugin-pug/blob/b7d26c7692b1146cb2717175a092d48bf9673e1e/src/index.ts#L147

The quotes will always be the other way around than JavaScripts configuration

Applelo:
No way to have singleQuote everywhere in my pug files ?

Shinigami92:
Currently no.
I have to inwestigate if this can be possible.
[...]

PR #312, opened 2021-11-24
Handle singleQuote option more precisely

shadowgate15:
This should now have code interpolation be using the pugSingleQuote unless it is inside of quotes.


shadowgate15:
#45

Shinigami92 (changes requested):
I will look into other changes later, I'm currently just on mobile.

Shinigami92 (changes requested):
Please revert `pnpm-lock.yaml`. It's unrelated to the PR

src/printer.ts
```diff
@@ -244,9 +244,8 @@ export class PugPrinter {
 		const wrapAttributesPattern: string = options.pugWrapAttributesPattern;
 		this.wrapAttributesPattern = wrapAttributesPattern ? new RegExp(wrapAttributesPattern) : null;
 
-		const codeSingleQuote: boolean = !options.pugSingleQuote;
 		this.codeInterpolationOptions = {
-			singleQuote: codeSingleQuote,
+			singleQuote: options.pugSingleQuote ?? options.singleQuote,
```
Shinigami92, line 248: This fallback is unnecessary cause it is already managed by `convergeOptions` in `src/options/converge.ts#L17` called in `src/index.ts#L91` and that is passed to the constructor of `PugPrinter`.

src/printer.ts
```diff
@@ -1572,11 +1574,12 @@ export class PugPrinter {
 	}
 
 	private each(token: EachToken): string {
+		const code: string = format(token.code, { parser: '__js_expression', ...this.codeInterpolationOptions });
```
Shinigami92, line 1577: Maybe we want to call `formatDelegatePrettier` here instead of doing it manually? 🤔

Shinigami92 (changes requested):
I think you should hugely enhance / extends the `tests/options/singleQuote/single-quote.test.ts` test cases. IMO the `expected` and `code` are not sufficient anymore now.

src/printer.ts
shadowgate15: I did that for consistence since the other options in that object do the same thing.

src/printer.ts
shadowgate15: or should it be `frameworkFormat`?

Shinigami92: You mean the name of the function? :thinking: 
I named it to delegate the formatting with/to prettier
It's not the framework but somewhat like an underlying/embedded format call

shadowgate15: like instead of `formatDelegatePrettier`. it doesn't look like `formatDelegatePrettier` actually takes in to account what the framework is.

Shinigami92: Have a look into L603 - L657 :slightly_smiling_face: 
it is only taking the requested formatter, the formatter could be a framework but also could be just e.g. json or css

shadowgate15: you could have something like `each item in "{{ array }}"` in angular for example.

Shinigami92: I would anyway suggest to not change it in this PR, cause it's unrelated :slightly_smiling_face:

shadowgate15: that was kind of what i was starting to think.

shadowgate15:
this is probably a pretty good start on this. there are still some issues that are related to pug-lexer and the way certain tokens are handled but I would say outside of the scope of this PR. Eventually I will try to circle back around to fix those up.

Ideally if I say `singleQuote: true` then all top level quotes should be ``` unless it is `"`"`.

src/printer.ts
```diff
@@ -582,11 +581,14 @@ export class PugPrinter {
...
 		val = val.trim();
+		const options: Options = { ...this.codeInterpolationOptions };
 		const wasQuoted: boolean = isQuoted(val);
 		if (wasQuoted) {
+			options.singleQuote = !this.options.pugSingleQuote;
 			val = val.slice(1, -1); // Remove quotes
 		}
-		val = format(val, { parser, ...this.codeInterpolationOptions });
+		val = format(val, { parser, ...options });
+		logger.debug('what is happening', val);
```
Shinigami92, line 591: :eyes: :laughing:

Shinigami92 (changes requested):

src/printer.ts
shadowgate15: haha I thought I got rid of that my bad.

Shinigami92:
I would really like that you extend the code fragments in `tests/options/singleQuote/single-quote.test.ts` so that still **every combination** of `singleQuote` + `pugSingleQuote` is covered

Shinigami92:
Will look into it tomorrow
Thanks for your work on this :heart:

Shinigami92 (changes requested):
I only have two more requests 🙂 Then I will merge it

1. You are allowed to remove `tests/options/singleQuote/single-quote.test.ts` due to it is covered by your new tests 🎉 
2. Please add in each of these `tests/options/singleQuote/**/*.pug` files a normal attribute to a pug token.
    So e.g. `div(:class='{ "pt-2": !dense }')` becomes `div(:class='{ "pt-2": !dense }', test='test')`
    Best would be that in the unformatted file always _wrong_ quote/format version is present, so it does explicitly get formatted by the test case
    This will help that no regression will happen there in the future

Shinigami92 (approved):

PR #314, opened 2021-11-30
Add `pugClassLocation` option

### LINKED ISSUE #313: Configurable class literal location
## Request / Idea
Have an option to set where class literals will be placed.

## Input

```pug
#class-after-attributes.class.class-2(type='text')

#class-before-attributes(type='text').class.class-2
```

## Expected Output

```pug
#class-after-attributes(type='text').class.class-2

#class-before-attributes.class.class-2(type='text')
```

shadowgate15:
closes #313

### DIFF
--- a/docs/guide/pug-specific-options.md
+++ b/docs/guide/pug-specific-options.md
@@ -432,7 +432,7 @@ foo(id="bar")
 
 ### Description
 
-Define where classes will be placed
+Define where classes will be placed.
 
 ### Options
 
@@ -441,7 +441,7 @@ Define where classes will be placed
 
 #### `'before-attributes'`
 
-Forces all valid class literals to be placed before attributes
+Forces all valid class literals to be placed before attributes.
 
 ```pug
 foo.bar.baz(attr='this-attr')
@@ -449,7 +449,7 @@ foo.bar.baz(attr='this-attr')
 
 #### `'after-attributes'`
 
-Forces all valid class literals to be placed after attributes
+Forces all valid class literals to be placed after attributes.
 
 ```pug
 foo(attr='this-attr').bar.baz
--- a/docs/guide/pug-specific-options.md
+++ b/docs/guide/pug-specific-options.md
@@ -426,6 +426,37 @@ foo(id="bar")
 
 ---
 
+## Class Location
+
+`pugClassLocation`
+
+### Description
+
+Define where classes will be placed
+
+### Options
+
+**Type:** `choice`
+**Default:** `'before-attributes'`
+
+#### `'before-attributes'`
+
+Forces all valid class literals to be placed before attributes
+
+```pug
+foo.bar.baz(attr='this-attr')
+```
+
+#### `'after-attributes'`
+
+Forces all valid class literals to be placed after attributes
+
+```pug
+foo(attr='this-attr').bar.baz
+```
+
+---
+
 ## Explicit div
 
 `pugExplicitDiv`
--- a/src/options/pug-class-location.ts
+++ b/src/options/pug-class-location.ts
@@ -1,0 +1,18 @@
+import type { ChoiceSupportOption } from 'prettier';
+import { CATEGORY_PUG } from '.';
+
+/** Pug class location. */
+export const PUG_CLASS_LOCATION: ChoiceSupportOption = {
+	since: '1.13.0',
+	category: CATEGORY_PUG,
+	type: 'choice',
+	default: 'before-attributes',
+	description: 'Define where classes be placed.',
+	choices: [
+		{ value: 'before-attributes', description: 'Forces all valid class literals to be placed before attributes.' },
+		{ value: 'after-attributes', description: 'Forces all valid class literals to be placed after attributes.' }
+	]
+};
+
+/** Pug class location. */
+export type PugClassLocation = 'before-attributes' | 'after-attributes';
--- a/src/options/pug-class-location.ts
+++ b/src/options/pug-class-location.ts
@@ -3,7 +3,7 @@ import { CATEGORY_PUG } from '.';
 
 /** Pug class location. */
 export const PUG_CLASS_LOCATION: ChoiceSupportOption = {
-	since: '1.13.0',
+	since: '1.19.0',
 	category: CATEGORY_PUG,
 	type: 'choice',
 	default: 'before-attributes',
--- a/src/printer.ts
+++ b/src/printer.ts
@@ -64,6 +64,7 @@ import type { PugEmptyAttributes, PugEmptyAttributesForceQuotes } from './option
 import { formatEmptyAttribute } from './options/empty-attributes/utils';
 import type { PugAttributeSeparator } from './options/pug-attribute-separator';
 import { resolvePugAttributeSeparatorOption } from './options/pug-attribute-separator';
+import { PugClassLocation } from './options/pug-class-location';
 import type { PugClassNotation } from './options/pug-class-notation';
 import type { PugCommentPreserveSpaces } from './options/pug-comment-preserve-spaces';
 import { formatPugCommentPreserveSpaces } from './options/pug-comment-preserve-spaces';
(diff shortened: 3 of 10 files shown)


shadowgate15:
I see the lint issues working on fixing them now.

src/options/pug-class-location.ts
```diff
@@ -0,0 +1,18 @@
+import type { ChoiceSupportOption } from 'prettier';
+import { CATEGORY_PUG } from '.';
+
+/** Pug class location. */
+export const PUG_CLASS_LOCATION: ChoiceSupportOption = {
+	since: '1.13.0',
```
Shinigami92, line 6:
```suggestion
	since: '1.19.0',
```

src/printer.ts
```diff
@@ -64,6 +64,7 @@ import type { PugEmptyAttributes, PugEmptyAttributesForceQuotes } from './option
 import { formatEmptyAttribute } from './options/empty-attributes/utils';
 import type { PugAttributeSeparator } from './options/pug-attribute-separator';
 import { resolvePugAttributeSeparatorOption } from './options/pug-attribute-separator';
+import { PugClassLocation } from './options/pug-class-location';
```
Shinigami92, line 67:
```suggestion
import type { PugClassLocation } from './options/pug-class-location';
```

tests/options/pugClassLocation/pug-class-location.test.ts
```diff
@@ -0,0 +1,56 @@
...
+				parser: 'pug',
+				plugins: [plugin],
+
+				pugClassLocation: 'before-attributes'
+			});
+
+			expect(actual).toBe(expected);
+		});
+		test('should position class literals after attributes', () => {
+			const expected: string = `(type="button", class=boolean ? "test" : "fact").class.class-before.after-class= "type"\n`;
```
Shinigami92, line 20: :eyes: does it really remove the `div` tag?
This doesn't look valid.

Shinigami92 (changes requested):
Will have a deeper look tomorrow, but looking good right now, good night

tests/options/pugClassLocation/pug-class-location.test.ts
shadowgate15: it does as the default.

Shinigami92: Using e.g. https://pughtml.com it gives me `unexpected token "start-attributes" - line 1:1`
so... this would be a bug :lady_beetle:

tests/options/pugClassLocation/pug-class-location.test.ts
```diff
@@ -0,0 +1,56 @@
+/* eslint-disable quotes */
+import { format } from 'prettier';
+import { plugin } from './../../../src/index';
+
+describe('Options', () => {
+	describe('pugClassLocation', () => {
```
Shinigami92, line 6: These tests are a good start, but IMO they need to be more complex.

We should also test for
1. wrapped attributes
2. indented tags

So I think you should outsource the pug content to pug files like in most other tests to have better multiline.
Also in e.g. VSCode it's possible to make a file compare between formatted and unformatted, that help sometimes to work with tests.

Shinigami92 (changes requested):

tests/options/pugClassLocation/pug-class-location.test.ts
shadowgate15: what is an example of wrapped attributes?

Shinigami92: `printWidth` ...

```pug
my-tag(
  my-attr="something"
)
```

tests/options/pugClassLocation/pug-class-location.test.ts
shadowgate15: this is actually general bug.
`(attr="test for no id or class") text` won't be reformatted correctly when `pugExplicitDiv` is set to `true` or `false`

shadowgate15: not sure that is something prettier should since it is just malformed pug anyways.

Shinigami92: I think you misunderstood something. I mean expected should be `const expected: string = `div(type="button", ...`
The test expectation is wrong here. A test should expect the correct thing not just the response it currently gives.
Then it fails and we need to fix the code until it response the correct result.

shadowgate15: I understood that. 
I'm saying that if the string contains `(attr="test for no id or class") text` it will not add the `div`.

Shinigami92: If it would be the `actual`, yes it would be invalid, but this line is the `expected`

```suggestion
			const expected: string = `div(type="button", class=boolean ? "test" : "fact").class.class-before.after-class= "type"\n`;
```

shadowgate15: I'm talking about in general, not this test example.

Shinigami92: pugExplicitDiv false only removes it if it's possible to omit it. so e.g. when using `.my-class(someattr="anything")` this is valid pug, with explicit div it would be `div.my-class(someattr="anything")`

shadowgate15: yes but if the line is incorrect to begin with it won't add the div.

Shinigami92: sure, because this is already false pug and therefore would not trigger my plugin printer at all

docs/guide/pug-specific-options.md
```diff
@@ -426,6 +426,37 @@ foo(id="bar")
 
 ---
 
+## Class Location
+
+`pugClassLocation`
+
+### Description
+
+Define where classes will be placed
```
Shinigami92, line 435:
```suggestion
Define where classes will be placed.
```

docs/guide/pug-specific-options.md
```diff
@@ -426,6 +426,37 @@ foo(id="bar")
...
+
+### Options
+
+**Type:** `choice`
+**Default:** `'before-attributes'`
+
+#### `'before-attributes'`
+
+Forces all valid class literals to be placed before attributes
```
Shinigami92, line 444:
```suggestion
Forces all valid class literals to be placed before attributes.
```

docs/guide/pug-specific-options.md
```diff
@@ -426,6 +426,37 @@ foo(id="bar")
...
+
+```pug
+foo.bar.baz(attr='this-attr')
+```
+
+#### `'after-attributes'`
+
+Forces all valid class literals to be placed after attributes
```
Shinigami92, line 452:
```suggestion
Forces all valid class literals to be placed after attributes.
```

Shinigami92 (approved):

Shinigami92:
Will merge it (potentially) tomorrow, it's currently to late at my location :sleeping:

tests/options/pugClassLocation/formatted-after.pug
```diff
@@ -0,0 +1,21 @@
+div(type="button", class=boolean ? "test" : "fact").before-class.class= "type"
+  div(type="button", class=boolean ? "test" : "fact").class-before.class.class-before.after-class= "type"
```
Shinigami92, line 2:
```suggestion
  div(type="button", class=boolean ? "test" : "fact").class.class-before.after-class= "type"
```

Somehow it seems that here the `class-before` would get duplicated. Could you please check and fix this?

Shinigami92 (changes requested):

tests/options/pugClassLocation/formatted-after.pug
shadowgate15: it is actually a straight up bug with the code and test. it was dropping a class on the top two divs. working on it now.

Shinigami92 (approved):

Shinigami92:
I will merge it now, but release could be a bit later, over weekend or so. I would like to manage some little other things before release 1.19 from my side.

PR #381, opened 2022-05-25
Use `node:` import prefix

Shinigami92:


Shinigami92:
Hey @antfu, I tried something out here: using an extra CI forced to NodeJS v14.6

> This version of pnpm requires at least Node.js v14.19

It looks like as we are using pnpm (same as in Vite) we are forced to at least that version :thinking: 
So should we bump the engine in Vite? (ref https://github.com/vitejs/vite/pull/8309)

image

Shinigami92:
I can use `fileURLToPath(new URL('./src', import.meta.url))` in `vite.config.ts` when switching to prettier v3

PR #390, opened 2022-06-09
Add support for frontmatter

jlangeveld:
First round - yaml frontmatter is being parsed en re-inserted.
Graymatter supports json out of the box, so I want to work that in too...


package.json
```diff
@@ -47,6 +47,7 @@
     "tsconfig.json"
   ],
   "dependencies": {
+    "gray-matter": "^4.0.3",
```
Shinigami92, line 50: Can we potentially use an optional dependency somehow here? Nearly every user that use my plugin don't use frontmatter. You are the first case in 2-3 years.
Adding a direct dependency would result in not used content for thousands of users.

tests/frontmatter/unformatted.pug
```diff
@@ -0,0 +1,4 @@
+---
+name: YAML test
```
Shinigami92, line 2: Maybe add a little bit more and unformatted yaml content, so we can also see that it gets formatted

Shinigami92 (changes requested):

tests/frontmatter/unformatted.pug
Shinigami92: Not resolved yet :wink: because you only added pug code, but not further _unformatted_ frontmatter code (e.g. you could add some whitespaces inside the values and add nested keys and lists)

package.json
Shinigami92: It brings sub-dependencies with it... https://github.com/jonschlinkert/gray-matter/blob/a5726b04f3167fadc764241deb545518c454eb82/package.json#L36-L41
And these have sub-sub-deps and so on.

fisker:
Maybe copy this https://github.com/prettier/prettier/blob/f2244bdc8621c0a001e3717f595bc8fd8653cce7/src/utils/front-matter/parse.js#L50? Or should Prettier expose it?

jlangeveld:
> Maybe copy this https://github.com/prettier/prettier/blob/f2244bdc8621c0a001e3717f595bc8fd8653cce7/src/utils/front-matter/parse.js#L50? Or should Prettier expose it?

Sorry for being off the radar, got a bit swamped...
I figured this as a better idea too.
At first just copy it, but ask prettier to expose it.

However, I got hung up trying to convince typescript to accept this javascript code without erorring out on missing type declarations.
Got any tips about that?

taniki:
As an 11ty user, I would be very interested by this plugin supporting pug with frontmatter.

@jlangeveld [parse.js](https://github.com/prettier/prettier/blob/9fbc9ce22df675a81a820780d47122a136da9061/src/main/front-matter/parse.js#L31) has been converted to typescript since then. Will this solve your problem?

I don't think I am proficient enough to pick up the ball.

taniki:
formatting frontmatter would be nice too.

Shinigami92:
@taniki I'm still here to support or lead

taniki:
@Shinigami92 I opened up a [new pull request](https://github.com/prettier/plugin-pug/pull/635). I don't know if there is a way for me to push directly into @jlangeveld branch.

I am a bit strained on time. It is not complete. I still have to figure out to import and format the frontmatter part whatever language it uses. This [file](https://github.com/prettier/prettier/blob/9fbc9ce22df675a81a820780d47122a136da9061/src/main/front-matter/embed.js) seems to be what we need.

PR #393, opened 2022-06-14
Fix pugClassNotation: 'as-is'

### LINKED ISSUE #392: Bug: Using pugClassNotation: 'as-is' reports Duplicate Attribute error in Vue/TS
## Info

| Plugin    | v2.0.0                  |
| Prettier  | v2.6.2                  |
| Framework | vue |
| Node      | v16                 |
| OS        | mac           |

## Prettier config

```
{
  plugins: [require.resolve('@prettier/plugin-pug')],

  singleQuote: true,
  trailingComma: 'all',
  semi: true,
  printWidth: 120,
  // prettier-ignore
  importOrder: [
    '^.*service.*$',
    '([./]|@/(.*)$)',
  ],
  importOrderSeparation: true,

  pugClassNotation: 'as-is',
  pugIdNotation: 'as-is',
  pugFramework: 'vue',
  pugSingleQuote: false,
  pugAttributeSeparator: 'none',
  pugSortAttributes: 'asc',
  pugSortAttributesBeginning: [
    '^v-if$',
    '^v-else-if$',
    '^v-else$',
    '^v-for$',
    '^v-model',
    '^:is$',
    '^:key$',
    '^v-on$',
    '^v-bind$',
    '^ref$',
    '^name$',
    '^:?type$',
    '^:value$',
    '^:?label$',
    '^:?src$',
  ],
  pugSortAttributesEnd: ['^:id', '^id', '^@click', '^@'],
}
```

## Input
```vue
<template lang="pug">
.sidebar(class="flex flex-col relative pb-28")
</template>
```

## Output or Error

```
Duplicate Attribute. vue(2)

```

## Additional Context
`pugClassNotation: 'as-is'`

Using Vue 3 with TypeScript and Tailwind.
The `class` attribute gets red underline and output in Problems console. The problem goes away if I remove the `.sidebar` class and put it inside of the class attribute instead, but we like to use class wrapper names for components. Code still compiles, just makes it look like we have lint issues.

The expected behavior would be that when setting is set to `as-is`, I can have classes in both places.

Shinigami92:
closes #392

### DIFF
--- a/tests/issues/issue-392/formatted.pug
+++ b/tests/issues/issue-392/formatted.pug
@@ -1,0 +1,1 @@
+.sidebar(class="flex flex-col relative pb-28")
--- a/tests/issues/issue-392/formatted.vue
+++ b/tests/issues/issue-392/formatted.vue
@@ -1,0 +1,3 @@
+<template lang="pug">
+.sidebar(class="flex flex-col relative pb-28")
+</template>
--- a/tests/issues/issue-392/issue-392.test.ts
+++ b/tests/issues/issue-392/issue-392.test.ts
@@ -1,0 +1,55 @@
+import { compareFiles } from 'tests/common';
+import { describe, expect, it } from 'vitest';
+
+describe('Issues', () => {
+  it('should handle pugClassNotation as-is', () => {
+    const { expected, actual } = compareFiles(__dirname, {
+      formatOptions: {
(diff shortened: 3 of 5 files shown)


Shinigami92:
@sfrendpax8 The tests do not report anything wrong yet :thinking: 
Are you sure you provided a fully working reproduction?

PR #407, opened 2022-07-25
Fix indent with pugExplicitDiv and pugClassNotation=attribute

### LINKED ISSUE #405: Bug: indentation issues with pugExplicitDiv and pugClassNotation=attribute
### Plugin Version

v2.1.1

### Prettier Version

v2.7.1

- [ ] none
- [X] vue
- [ ] angular
- [ ] svelte

### Node Version

v16.16.0

- [X] Linux
- [ ] macOS
- [ ] Windows

### Prettier config

```json
{
  "trailingComma": "es5",
  "tabWidth": 2,
  "semi": false,
  "singleQuote": true,
  "pugSingleQuote": false,
  "pugWrapAttributesThreshold": 1,
  "pugClassNotation": "attribute",
  "pugExplicitDiv": true
}
```

### Input

```pug
<template lang="pug">
.test
  .test2
    .test3
      .test4
</template>
```

### Output or Error

```bash
div(class="test")
div(class="test2")
  div(class="test3")
    div
```

### Expected Output

```pug
div(class="test")
  div(class="test2")
    div(class="test3")
      div
```

### Additional Context

When adding `"pugClassNotation": "attribute"` option and then reformatting with prettier, the indentation changed.
(In the given example, the second div indentation). 

However, if i first add explicit div like that manually, it works as expected:

```
div.test
  div.test2
    div.test3
      div.test4
```

Shouldn't  `"pugClassNotation": "attribute"` option enforce ` "pugExplicitDiv": true` and apply it before re-indenting?
I think this is not correct in pug :

```(class="test")```

but this what I get if I remove ` "pugExplicitDiv": true` option : 

Before:
```.test```

After:
```(class="test")```

### PR DESCRIPTION
**Pull request by Shinigami92**
closes #405

### DIFF
--- a/src/printer.ts
+++ b/src/printer.ts
@@ -1238,14 +1238,16 @@ export class PugPrinter {
       }
       if (
         this.nextToken &&
-        ['text', 'newline', 'indent', 'eos'].includes(this.nextToken?.type)
+        ['text', 'newline', 'indent', 'outdent', 'eos'].includes(
+          this.nextToken.type,
+        )
       ) {
         const classes: string[] = this.classLiteralToAttribute.splice(
           0,
           this.classLiteralToAttribute.length,
         );
         this.result += `(class=${this.quoteString(classes.join(' '))})`;
-        if (this.nextToken?.type === 'text') {
+        if (this.nextToken.type === 'text') {
           this.result += ' ';
         }
       }
--- a/src/printer.ts
+++ b/src/printer.ts
@@ -1235,6 +1235,13 @@ export class PugPrinter {
         this.previousToken?.type !== 'class'
       ) {
         this.result += 'div';
+        if (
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/printer.ts [hunk 1]
```diff
@@ -1235,17 +1235,26 @@ export class PugPrinter {
         this.previousToken?.type !== 'class'
       ) {
         this.result += 'div';
+        if (
+          this.indentLevel === 0 &&
+          this.previousToken?.type !== 'newline' &&
+          this.previousToken?.type !== 'outdent'
+        ) {
+          this.indentLevel++;
+        }
```
Shinigami92, line 1244: Not sure if this is the correct way 😕

Shinigami92: Thx to @ST-DDT we found the correct way of handling it :partying_face:

PR #423, opened 2022-12-06
Add semi code interpolation option

eevleevs:
Interpolated code currently does not respect the *semi* option.

With `semi: false`

Before this change
```js
script.
  console.log("test");
```

After this change
```js
script.
  console.log("test")
```

### DIFF
--- a/src/printer.ts
+++ b/src/printer.ts
@@ -197,7 +197,7 @@ export class PugPrinter {
 
   private readonly codeInterpolationOptions: Pick<
     RequiredOptions,
-    | 'semi',
+    | 'semi'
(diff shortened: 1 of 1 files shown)


OiYouYeahYou:
@eevleevs have you had a chance to look at the type error from CI?

eevleevs:
Unclear to me, please help.

OiYouYeahYou:
@Shinigami92, I've confirmed this works as described both before and after changes. And have added some tests on #426.

I am curious why the printer doesn't pass the options in full to the interpolation?

Shinigami92:
closed via #426

PR #427, opened 2023-01-14
Fix trailing whitespace after tag

### LINKED ISSUE #425: Bug: Trailing whitespace in tags is not trimmed
### Plugin Version

v2.3.0

### Prettier Version

v2.8.1

- [X] none
- [ ] vue
- [ ] angular
- [ ] svelte

### Node Version

v16.17.0

- [ ] Linux
- [X] macOS
- [ ] Windows

### Prettier config

```json
{

}
```

### Input

```pug
html
  body
    p 
      a Link
```

### Output or Error

```bash
html
  body
    p 
      |
      a Link
```

### Expected Output

```pug
html
  body
    p
      a Link
```

### Additional Context

Note the mixed whitespace immediately following the `p` element.
The major problem with this bug is that running prettier on this output yields an even different result (more pipes).

Shinigami92:
closes #425

### DIFF
--- a/src/printer.ts
+++ b/src/printer.ts
@@ -1481,7 +1481,7 @@ export class PugPrinter {
     let val: string = token.val;
     let needsTrailingWhitespace: boolean = false;
 
-    const endsWithWhitespace: boolean = val[val.length - 1] === ' ';
+    let endsWithWhitespace: boolean = val[val.length - 1] === ' ';
 
     if (this.pipelessText) {
       switch (this.previousToken?.type) {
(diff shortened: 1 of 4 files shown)


src/printer.ts
```diff
@@ -1567,6 +1567,10 @@ export class PugPrinter {
       val = ` ${val}`;
     }
 
+    if (val === ' ' && this.nextToken?.type === 'indent') {
+      val = val.trimEnd();
+    }
```
Shinigami92, line 1572: Awaiting to find a time slot with @ST-DDT to discuss if this is the optimal solution

Shinigami92: Thx to @ST-DDT we found that the prev applied fix was even incorrect :pray: 
Now it does what the actual expected result should be + it's optimized

ST-DDT (approved):

PR #457, opened 2023-07-23
Update Changelog

Shinigami92:
@lehni @fisker Please confirm that this Changelog is correct and I will release a v2.5.2

You can also see here https://github.com/prettier/plugin-pug/compare/2.5.1...main what I changed since v2.5.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
 # Next
 
-[diff](https://github.com/prettier/plugin-pug/compare/2.5.1...main)
+[diff](https://github.com/prettier/plugin-pug/compare/2.5.2...main)
+
+# 2.5.2
+
+[diff](https://github.com/prettier/plugin-pug/compare/2.5.1...2.5.2)
+
+This version does not contain any changes to the plugin itself, but only internal changes to the docs, build and CI.
+
+This might be the last version that supports Prettier v2 as well as CJS.  
+The next version will probably the v3 release and will only support Prettier v3 and ESM.
 
 # 2.5.1
 


lehni:
@Shinigami92 LGTM!

lehni (approved):

Shinigami92:
> LGTM!

Thx that's valuable to me, but I still need the :heavy_check_mark:-Approval also from @fisker, because he is one of the Prettier Core Maintainer
I also asked the "is migration to ESM-only"-safe-way in the Prettier Core Discord Server and maybe some other maintainers can also approve there

fisker (approved):

Shinigami92:
@lehni @fisker https://www.npmjs.com/package/@prettier/plugin-pug/v/2.5.2 is released

Not sure if I found today time anymore to start working on the ESM rewrite
I will first switch some stuff (like set package.json type to module)
then I will look into the `await format` changes

Just give me some time and you don't need to work so much yourself on the PRs, because at the end I need to merge it anyways ^^
I just need to have the time

lehni:
@Shinigami92 I don't think the ESM conversion is technically needed right away to make the plugin work with prettier v3. I tested #456 with `yarn link` on my projects and it worked as expected.

Shinigami92:
> @Shinigami92 I don't think the ESM conversion is technically needed right away to make the plugin work with prettier v3. I tested #456 with `yarn link` on my projects and it worked as expected.

It is less a "need" but more a "like"
I'm a core member in @vitejs Team and we are trying now to bring the ecosystem forward with ESM
I also don't want to use a CJS+ESM for this plugin, as this would double the unpacked package size (not so big for this plugin, but in @faker-js it is painful)

Aside that, ESM is the future and some day I would need to do this step anyway, so why not now 🙂

lehni:
I'm not arguing against ESM, absolutely embracing it myself in my projects. I'm just saying there's a viable quick path forward to making this plugin work with prettier 3, which would maybe remove some of the stresses of having to do the ESM transition at the same time.

Shinigami92:
@lehni 
The stress is not coming from the project/plugin, but from outside
So it is more a "I need to find time-slots" to work on the project at all

PR #509, opened 2024-09-08
Multiple fixes for class attribute conversions

### LINKED ISSUE #508: Bug: pugClassNotation: 'attribute' outputs junk if class literals are after attributes
### Plugin Version

3.1.0

### Prettier Version

3.3.3

- [ ] none
- [ ] vue
- [ ] angular
- [ ] svelte

### Node Version

20.17.0

- [ ] Linux
- [X] macOS
- [ ] Windows

### Prettier config

```json
{
  pugClassNotation: 'attribute'
}
```

### Input

```pug
div.foo
div(attribute="value").foo
div(attribute="value" class="bar").foo
div(attribute="value" class="bar").foo.baz
div(
  attribute="value"
  class="bar"
).foo.baz
div.foo(attribute="value" class="bar").baz
div: span.foo
div: span(class="foo")
div.foo: span.bar(class="baz").foo
div.foo(class="bar").baz: h1.foo(class="bar").baz Let's really #[span.foo(class="bar").baz get nasty]!
```

### Output or Error

```bash
(class="foo")
div(attribute="value")div(class="foo")
div(attribute="value", class="bar")div(class="foo")
div(attribute="value", class="bar")div(class="foo baz")
div(attribute="value", class="bar")div(class="foo baz")
(attribute="value", class="foo bar")div(class="baz")
div: span(class="foo") // <- this line was correct!
div: span(class="foo") // <- this line was correct!
div: span(class="foo")
: span(class="foo bar baz")div(class="foo")
(class="foo bar")div: h1(class="baz foo bar")div(class="baz") Let's really #[span(class="foo bar")div(class="baz") get nasty]!
```

### Expected Output

```pug
div(class="foo")
div(attribute="value", class="foo")
div(attribute="value")
div(attribute="value", class="bar foo")
div(attribute="value", class="bar foo baz")
div(attribute="value", class="bar foo baz")
div(attribute="value", class="bar foo baz")
div: span(class="foo")
div: span(class="foo")
div(class="foo"): span(class="bar baz foo")
div(class="foo bar baz"): h1(class="foo bar baz") Let's really #[span(class="foo bar baz") get nasty]!
```

### Additional Context

My workaround for this issue was to first use the option that refactors all class literals to the beginning and only then enable this class attribute option.

This is easy to repro by expanding the `pugClassNotation` test with the above code.

The offending code is easy to spot at the beginning of the private `class` function in `printer.ts,` but it's not immediately clear to me why that code works in the first place (do multiple chained attributes get merged by some follow-up logic?) or how to fix it properly. I'll open a PR if I can puzzle it out 🤷‍♂️

My verbosely commented version of the code that needs fixing:
```
    // Special handling for class literals when using the attribute notation.
    if (this.options.pugClassNotation === 'attribute') {
      // Add the class to the list of class literals to be converted to attributes.
      this.classLiteralsToBeConvertedToAttributes.push(token.val);

      // TODO: This can print junk (bug) if the class literal is after attributes. Investigate.
      if (
        this.previousToken?.type !== 'tag' &&
        this.previousToken?.type !== 'class'
      ) {
        this.result += `${this.computedIndent}div`;
      }

      // If this is the last class token (i.e. we are done with the literals)...
      // TODO: This probably does not handle inline elements chained with : correctly. Investigate.
      if (
        this.nextToken &&
        ['text', 'newline', 'indent', 'outdent', 'eos'].includes(
          this.nextToken.type,
        )
      ) {
        // Take a copy of the class literals to be converted to attributes and clear the list.
        const classes: string[] =
          this.classLiteralsToBeConvertedToAttributes.splice(
            0,
            this.classLiteralsToBeConvertedToAttributes.length,
          );

        // Print all the class literals as one class attribute.
        // TODO: What happens if the class attribute is already present? Investigate.
        this.result += `(class=${this.quoteString(classes.join(' '))})`;
        if (this.nextToken.type === 'text') {
          this.result += ' ';
        }
      }

      // Done processing the class token. Return early.
      return;
    }
```

MrHaila:
Closes #508 

This PR adds tests and fixes for previously untested variations of class literals in increasingly complex scenarios. Most of them output non-valid Pug code currently.

I fixed all the new test cases to output valid Pug code without regressions, so this is probably safe to merge. I don't have full confidence that there are no other (currently untested) bad combinations of the various options, but this is for sure better than before.

Examples of code that previously outputted invalid Pug but now work:

```pug
// Class literals at the end of attributes did not work.
div(attribute="value").foo
div(class="bar", attribute="value").foo
div(attribute="value", class="bar").foo.baz
div(
  attribute="value"
  class="bar"
).foo.baz
div.foo(attribute="value" class="bar").baz

// Class literals ending at : did not work
div(class="foo").bar: span Text
div: span.foo
div: span(class="foo")
div.foo: span.bar(class="baz").foo

// Class literals in an interpolated line did not work
div.foo(class="bar").baz: h1.foo(class="bar").baz Let's really #[span.foo(class="bar").baz get nasty]!
```

### DIFF
--- a/src/printer.ts
+++ b/src/printer.ts
@@ -1277,12 +1277,15 @@ export class PugPrinter {
 
       // An extra div should be printed if...
       if (
+        this.previousToken === undefined ||
         // ...the previous token indicates that this was the first class literal and thus a div did not previously exist...
-        (this.previousToken?.type !== 'tag' &&
-          this.previousToken?.type !== 'class' &&
-          this.previousToken?.type !== 'end-attributes') ||
+        this.checkTokenType(
+          this.previousToken,
+          ['tag', 'class', 'end-attributes'],
+          true,
+        ) ||
         // ...OR the previous token is a div that will be removed because of the no explicit divs rule.
-        (this.previousToken?.type === 'tag' &&
+        (this.previousToken.type === 'tag' &&
           this.previousToken.val === 'div' &&
           this.nextToken?.type !== 'attribute' &&
           !this.options.pugExplicitDiv)
@@ -1291,10 +1294,14 @@ export class PugPrinter {
       }
 
       if (
-        this.nextToken &&
-        ['text', 'newline', 'indent', 'outdent', 'eos', ':'].includes(
-          this.nextToken.type,
-        )
+        this.checkTokenType(this.nextToken, [
+          'text',
+          'newline',
+          'indent',
+          'outdent',
+          'eos',
+          ':',
+        ])
       ) {
         // Copy and clear the class literals list.
         const classes: string[] = this.classLiteralToAttribute.splice(
@@ -1333,7 +1340,7 @@ export class PugPrinter {
           this.result += `(class=${this.quoteString(classes.join(' '))})`;
         }
 
-        if (this.nextToken.type === 'text') {
+        if (this.nextToken?.type === 'text') {
           this.result += ' ';
         }
       }
--- a/src/printer.ts
+++ b/src/printer.ts
@@ -1290,24 +1290,6 @@ export class PugPrinter {
         this.result += `${this.computedIndent}div`;
       }
 
-      // If last token was the end of attributes...
-      // if (this.previousToken?.type === 'end-attributes') {
-      //   // ...and the attributes contained a class attribute...
-      //   const lastClassIndex: number = this.tokens
-      //     .slice(0, this.currentIndex)
-      //     .findIndex((t) => t.type === 'attribute' && t.name === 'class');
-
-      //   if (lastClassIndex > -1) {
-      //     // ...then find the last class attribute and insert the new class into it.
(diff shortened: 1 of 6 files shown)


MrHaila:
This PR is ready for review and would close #508 if deemed merge-worthy.

Shinigami92:
Back from vacation, I'll have a look the next few days, because I need to ensure that everything is correct

Shinigami92:
Added 1-level indentation tests and made use of `checkTokenType`

I just saw that for some cases we have still a `div` leftover like with [L17](https://github.com/prettier/plugin-pug/pull/509/files#diff-25e0a84970cc1f8825f5ba0c97ad0f91be510bb1876753ca5d79df15a0dbebd2R17)
But either this might get formatted by just a second run, or it can be fixed by a later MR (if it is even a problem)

PR #549, opened 2025-02-08
Fix issue 547

### LINKED ISSUE #547: Bug: adds unwanted space if pugSingleFileComponentIndentation is true
### Plugin Version

v3.2.1

### Prettier Version

v3.4.2

- [ ] none
- [x] vue
- [ ] angular
- [ ] svelte

### Node Version

v23.5.0

- [ ] Linux
- [ ] macOS
- [x] Windows

### Prettier config

```json
{
  "pugSingleFileComponentIndentation": true
}
```

### Input

```pug
<template lang="pug">
ul
  each val in [1, 2, 3, 4, 5]
    li= val
</template>
```

### Output or Error

```bash
<template lang="pug">
  ul
    each val in [1, 2, 3, 4, 5]
      li  = val
</template>
```

### Expected Output

```pug
<template lang="pug">
  ul
    each val in [1, 2, 3, 4, 5]
      li= val
</template>
```

### Additional Context

_No response_

Shinigami92:
closes #547 

- #547

### DIFF
--- a/src/printer.ts
+++ b/src/printer.ts
@@ -1793,7 +1793,8 @@ export class PugPrinter {
   }
 
   private async code(token: CodeToken): Promise<string> {
-    let result: string = this.computedIndent;
+    let result: string =
+      this.previousToken?.type === 'tag' ? '' : this.computedIndent;
     if (!token.mustEscape && token.buffer) {
       result += '!';
     }
--- a/tests/issues/issue-547/formatted.vue
+++ b/tests/issues/issue-547/formatted.vue
@@ -1,0 +1,5 @@
+<template lang="pug">
+  ul
+    each val in [1, 2, 3, 4, 5]
+      li= val
+</template>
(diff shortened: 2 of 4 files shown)


Shinigami92:
@Ralac07 please review and tell me if you have other more complex test cases I can add

Ralac07:
I just used the sample code on the pug website, tests should probably include those but I'm not really sure
https://pugjs.org/language/iteration.html
sorry for my delay in responding

Shinigami92:
> I just used the sample code on the pug website, tests should probably include those but I'm not really sure [pugjs.org/language/iteration.html](https://pugjs.org/language/iteration.html) sorry for my delay in responding

That's a really good one. But do you have any more cases from your real world? I'm asking because I would like to now if I only need to consider `tag` as a previous token, or if there is more.

https://github.com/prettier/plugin-pug/blob/cc7a0b60467e365287cfeace8d47f0d7b998fe3b/src/printer.ts#L1797

Ralac07:
I don't use pug much, sorry.

Shinigami92:
So if I understood you correctly, you are not affected in production code by this?
If that's the case, I would set this right now to low prio on my task-list and await more feedback from other users and if they encounter this bug.
Because since the repo was started, this bug was never reported ^^

PR #418, opened 2022-10-21
Transform trailing whitespace to recommended way

jwerre:
Created test to deal with issue #417 

The following pug:

```pug
p Aliquam pulvinar enim eu enim finibus euismod. 
	| a(href="https://example.com") Link
	|  Suspendisse suscipit nunc non mauris sodales elementum.
```

should reformat to:

```
p Aliquam pulvinar enim eu enim finibus euismod.
	|
	| a(href="https://example.com") Link
	|
	| Suspendisse suscipit nunc non mauris sodales elementum.
```

PR #567, opened 2025-03-27
Fix whitespace handling for interpolation within piped text

### LINKED ISSUE #486: Bug: Tag interpolation not formatted correctly within piped text
## Info

| Plugin    | v3.0.0                  |
| Prettier  | v3.2.5                  |
| Framework | Vue |
| Node      | v18.19.0                  |
| OS        | Linux           |

## Prettier config

```json
{
    "endOfLine": "auto",
    "pugCommentPreserveSpaces": "keep-leading",
    "overrides": [
        {
            "files": "*.md",
            "options": {
                "tabWidth": 2
            }
        }
    ]
}
```

## Input

```pug
div
    | #[strong foo] bar
    | #[strong foo] bar
```

## Output or Error

```pug
div
    |#[strong foo] bar
    |         #[strong foo] bar
```

## Expected Output

```pug
div
    | #[strong foo] bar
    | #[strong foo] bar
```

alexis-aquanty:
Fixes #486 

This also reverts the unintuitive behaviour introduced by #418. While this behaviour may be desirable for some people, I would argue that it's much more of a footgun than anything. It is only useful in the case when the user intentionally adds extra whitespace at the beginning of a line of piped text and wants the extra space to be preserved (which is explicitly [not recommended](https://pugjs.org/language/plain-text.html#not-recommended) by Pug), but does not know that they should use an empty piped line instead, or alternatively use tag interpolation, which IMO is much more intuitive. On the other hand, it is very common to accidentally add more whitespace than intended at the beginning of a line of piped text, especially when copying and pasting, in which case this behaviour is both unintuitive and undesirable.


Shinigami92:
I think this is a big breaking change, is it? Not in terms of how spaces are handled, but for the plugin consumer's expectation and opinionated formatting style 🤔 

Should we consider a new option so the plugin consumer can decide what to prefer?
Maybe this is something like `preserveWhitespace`

alexis-aquanty:
True, an option would make sense for this. I'll look into adding one.

tests/issues/issue-417/formatted.pug
Shinigami92, line 1: **todo (blocking):** please restore this issue-417 associated test, so it is still tracked and ensures backwards compatibility

it's fine if the test is somewhat duplicated inside options/pugPreserveWhitespace

tests/issues/issue-417/issue-417.test.ts
```diff
@@ -3,7 +3,9 @@ import { describe, expect, it } from 'vitest';
 
 describe('Issues', () => {
   it('should add two empty piped lines before and after link tag', async () => {
-    const { expected, actual } = await compareFiles(import.meta.url);
+    const { expected, actual } = await compareFiles(import.meta.url, {
+      formatOptions: { pugPreserveWhitespace: true },
```
Shinigami92, line 7: **issue:** I think this is wrong here, the test should not be touched at all and the default for `pugPreserveWhitespace` should be set, so that the test succeeds without touching this test. The `tests/options/pugPreserveWhitespace/pug-preserve-whitespace.test.ts` is then testing all the different cases: _default_, `false` and `true`

Shinigami92:
you can fix all lint and formatting errors normally by running `pnpm run preflight` locally

tests/issues/issue-417/issue-417.test.ts
alexis-aquanty: I understand wanting to preserve backwards compatibility, but I think that making `pugPreserveWhitespace` default to `true` would be undesirable for the reasons I explained in the PR description. This behaviour seems far more likely to trip people up than to be helpful. If you insist, I will change it, but it doesn't seem like a good default to me.

Shinigami92: We can consider to change the default in a breaking v4 change in the future (you can also add a comment into the code for that!).
But for now lets use a default compatible to the version.
I use TODO comments like `// TODO @Shinigami92 2025-03-31: ...`

And you are the first person after nearly 6 years of wanting this behavior 😜

src/options/pug-preserve-whitespace.ts
```diff
@@ -0,0 +1,12 @@
+import type { BooleanSupportOption } from 'prettier';
+import { CATEGORY_PUG } from './constants';
+
+/** Pug preserve whitespace. */
+export const PUG_PRESERVE_WHITESPACE: BooleanSupportOption = {
+  // since: '3.3.0',
+  category: CATEGORY_PUG,
+  type: 'boolean',
+  // TODO @Shinigami92 2025-03-31: considering defaulting to false
```
Shinigami92: **chore:** 
```suggestion
  // TODO @alexis-aquanty 2025-03-31: considering defaulting to false in v4
```

Shinigami92 (approved):

Issue #512, opened 2024-10-01
Enhance Variable Formatting: Convert Double Quotes to Single Quotes


Anion11:
## Request / Idea
I would like to be able to correct double quotes to single quotes in variables (specifically in variables, not attributes)

## Input

```pug
const variable = "test"
```
or
```pug
+mixin-name({ text: "test" })
```
## Expected Output

```pug
const variable = 'test'
```
or
```pug
+mixin-name({ text: 'test' })
```
Maybe there is already such a function and I'm doing something wrong?

My .prettierrc:
```.prettierrc
{
  "plugins": ["@prettier/plugin-pug"],
  "printWidth": 100,
  "tabWidth": 2,
  "useTabs": false,
  "semi": false,
  "singleQuote": true,
  "quoteProps": "as-needed",
  "trailingComma": "none",
  "bracketSpacing": true,
  "bracketSameLine": false,
  "arrowParens": "avoid",
  "htmlWhitespaceSensitivity": "css",
  "singleAttributePerLine": true,
  "pugIdNotation": "as-is",
  "pugAttributeSeparator": "none",
}

```

Shinigami92:
Did you tried all combinations with puSingleQuotes?

Anion11:
yes, pugSingleQuote doesn't help either

PR #578, opened 2025-04-12
Feat #512: Enable double/single quote conversion on various pug statements

DavidSemke:
### Overview
This PR enables double to single quote conversion and vice versa for strings associated with these tokens:
- case
- when
- if
- else if
- unless
- each-in
- each-of
- while
- mixin definition
- mixin call
- &attributes

So when `singleQuote` or `pugSingleQuote` is false, this:
```
//- Case statement
case 'red'
	when 'red'
		p It is red
	default
		p It is not red

//- Conditionals
if 'red' === '\'red\''
	p Red is red
else if 'blue' === '"blue"'
	p Blue is blue

unless 'blue' !== "\"blue\""
	p Blue is blue

//- Iteration
each val in ['One', "Two"]
	p= val

while 'red' === "'blue'"
	- break

//- Mixin def
mixin element(el='button')
	p= el

//- Mixin call
+element('button')

//- &attributes
div&attributes({x: 'x', y: 'y'})
```
becomes this:
```
//- Case statement
case "red"
  when "red"
    p It is red
  default
    p It is not red

//- Conditionals
if "red" === "'red'"
  p Red is red
else if "blue" === '"blue"'
  p Blue is blue

unless "blue" !== '"blue"'
  p Blue is blue

//- Iteration
each val in ["One", "Two"]
  p= val

while "red" === "'blue'"
  - break

//- Mixin def
mixin element(el = "button")
  p= el

//- Mixin call
+element("button")

//- &attributes
div&attributes({ x: "x", y: "y" })
```
Prettier is used to format, so some spacing is added as well in some cases.

### In Addition
I consolidated quote formatting tests under tests/options/singleQuote. Folder tests/quotes was deleted due to redundancy (all unique test cases offered at this location were preserved). I also moved:
- singleQuote/single-false-pug-false tests to singleQuote/single-false
- singleQuote/single-true-pug-true tests to singleQuote/single-true

This is because the formatting in these cases is equivalent.


Shinigami92:
This is a very huge PR with 42 file changes. Please do not remove any existing test files, because otherwise it is hard for me as a maintainer to ensure that there are no breaking changes.
If there are breaking changes, we need to provide feature flags / options.

DavidSemke:
All removed test files have been restored and updated to reflect the new quote conversions.

Shinigami92 (approved):

Shinigami92:
@DavidSemke @Anion11 anything you want to do before I hit merge and release a next minor?

PR #601, opened 2025-08-11
Handle unexpected missing block SyntaxError when parsing unbuffered code

sderrow:
Fixes #600.

PR #609, opened 2025-08-11
Handle else blocks and comments in unbuffered code

sderrow:
In #601, I missed a couple critical cases:
- `else` and `else if` statements, where the syntax error is actually at the start of the line, not the end
- Lines that have comments at the end  (e.g., `if (foo) // here, we're doing xyz`), since adding an empty block at the end would mean it'd be commented out if it's a `//` comment

@Shinigami92 hope you don't mind that I went straight for a PR instead of opening an issue first.

### DIFF
--- a/src/printer.ts
+++ b/src/printer.ts
@@ -1823,13 +1823,24 @@ export class PugPrinter {
     return val.slice(0, -1);
   }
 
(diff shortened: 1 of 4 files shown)


Shinigami92 (approved):

Shinigami92:
As usual, release hopefully in the evening, because I wont like to release from my company laptop 😉

PR #635, opened 2026-01-14
Add support for frontmatter

taniki:
- follow-up of #390
- rebase to catch-up with current HEAD
- copy frontmatter parsing and printing from prettier source code
- :warning: it works but doesn't pass the test because of formatting the frontmatter code. Currently digging into that.

Inspiration: https://github.com/LukeHagar/prettier-plugin-openapi/tree/b1a2f010724eb0b29d0e02ed56b3e493b4ec1438/src/prettier-markdown


src/utils/frontmatter/replace-non-line-breaks-with-space.js
```diff
@@ -0,0 +1,19 @@
+// import * as assert from "#universal/assert";
```
taniki, line 1: I have no idea how to put it back

package.json
```diff
@@ -62,7 +62,8 @@
     "tsconfig.json"
   ],
   "dependencies": {
-    "pug-lexer": "^5.0.1"
+    "gray-matter": "^4.0.3",
```
Shinigami92: **issue (blocking):** As with https://github.com/prettier/plugin-pug/pull/390#discussion_r893524592, this is a blocking thing for plugin-pug, as very few consumers are using frontmatter. Therefore we need a way to allow this as optional dependency and load it only if necessary.


... however actually I now saw that gray-matter is not used anymore, is it? 🤔 If that works that would be awesome 🚀

Shinigami92 (changes requested):

package.json
taniki: yes but npm and pnpm are very obscure to me. Because of that I haven't committed changes in `pnpm-lock.yaml` and `package.json` yet.

Shinigami92: When you have installed pnpm (either directly or via corepack) you might run `pnpm run preflight` always from the project root folder, and it will do all necessary things like formatting, linting, testing and generate the lock file.

taniki:
@Shinigami92 I am trying to use https://github.com/inogai/prettier/tree/main/src/language-markdown as a reference but the structure seems a bit different from yours.

package.json
taniki: oh I did a `npx pnpm rm -S gray-matter` to solve this. should I revert and run the preflight command?

Shinigami92: I think you could revert the full `package.json` and `pnpm-lock.yaml` to the state of main branch and then run `pnpm run preflight`

Shinigami92:
> @Shinigami92 I am trying to use [inogai/prettier@`main`/src/language-markdown](https://github.com/inogai/prettier/tree/main/src/language-markdown?rgh-link-date=2026-01-14T11%3A18%3A52Z) as a reference but the structure seems a bit different from yours.

Yeah, many plugins are fully driven by their own owners like myself. It is fully up to us maintainers how to run our plugins, but we can ship the plugin as officially under the `@prettier` org / npm-scope as a sign of trust.

taniki:
frontmatter support in markdown, css, and html seems to be done through support of embeds:

```

const printer = {
  preprocess,
  print: genericPrint,
  embed,
  massageAstNode: clean,
  hasPrettierIgnore,
  insertPragma,
  getVisitorKeys,
};

export default printer;
```

but I can't find such mecanism in yours. should insert something like inserting a case in the `switch` section?

Shinigami92:
> frontmatter support in markdown, css, and html seems to be done through support of embeds:
> 
> ```
> 
> const printer = {
>   preprocess,
>   print: genericPrint,
>   embed,
>   massageAstNode: clean,
>   hasPrettierIgnore,
>   insertPragma,
>   getVisitorKeys,
> };
> 
> export default printer;
> ```
> 
> but I can't find such mecanism in yours. should insert something like inserting a case in the `switch` section?

Yeah... that might be a huge problem, because I implemented plugin-pug when a lot of such things where not available or stable, and then also as pug is a "language" which is very easily written from left to write, top to bottom, it was far more easy to just process the token stream as an array / iterator.

You could have a look how embedded js or css is formatted right now as well as some expressions like embedded vue or angular syntax.
From that you might get a picture of how to achieve similar thing for frontmatter.

src/utils/frontmatter/parse.js
Shinigami92, line 1: **issue (blocking):** CLI is telling that these files are not TypeScript files, but JavaScript. You need to use TypeScript.

Shinigami92 (changes requested):

src/utils/frontmatter/parse.js
taniki: That will be too much for me right now. Thanks for the support.

src/printer.ts
```diff
@@ -478,6 +492,17 @@ export class PugPrinter {
     }
   }
 
+  private async frontmatterFormat(frontmatter: string): Promise<string> {
+    const options: Options = {
+      parser: 'yaml',
+      collectionStyle: 'block',
```
taniki, line 499: this should enforce formatting of flow style array and dict but it doesn't and I don't know why.

taniki:
I am hitting a wall here.

- I have no clue about how to efficiently convert the imported javascript code into typescript.
- the build-in yaml formatter doesn't seem to acknowledge the formatting option enforcing block-style list and dictionaries.
- there are some lint errors beyond my understanding.

All the bricks are in place tho. If someone to take it up from there, it will be great.

PR #655, opened 2026-09-14
Fix multiline mixin call indentation

### LINKED ISSUE #620: Bug: mixin formating breaks
When formating, mixins arguments breaks. This is related to this issue #66. But it can be fixed without waiting for dead pugjs update. My suggestion is to ignore any of formating rules for mixins. It is not a big deal to format js code manually, I do this all time and it is not too much code. In exchange we will get all other formating rules. Better to not have some feature then to tave it broken.

## Suggestion

There is patched version: [plugin-pug-formatting-patched.zip](https://github.com/user-attachments/files/22990669/plugin-pug-formatting-patched.zip)

https://github.com/prettier/plugin-pug/blob/33c86341f0addf48b09968f2545efcf18331d046/src/printer.ts#L2186-L2207

```diff
private async call(token: CallToken): Promise<string> {
     let result: string = `${this.computedIndent}+${token.val}`;
     let args: string | null = token.args;
-    if (args) {
-      args = args.trim().replaceAll(/\s\s+/g, ' ');
-      // Place an x at the beginning to preserve brackets,
-      // then remove the x after format.
-      args = await format(`x(${args})`, {
-        parser: 'babel',
-        ...this.codeInterpolationOptions,
-        semi: false,
-      });
-      args = args.trim().slice(1);
-      result += args;
-    }
+    result += `(${args})`;
+    
     this.currentLineLength += result.length;
     this.possibleIdPosition = this.result.length + result.length;
     this.possibleClassPosition = this.result.length + result.length;
     return result;
   }
```

## Info

| Plugin    | ^3.4.2                  |
| Prettier  | ^3.6.2                  |
| Framework | none |
| Node      | v22.14.0                  |
| OS        | win           |

## Prettier config

```json
{
  "pugCommentPreserveSpaces": "trim-all",
  "pugSortAttributes": "as-is",
  "pugWrapAttributesPattern": "",
  "pugWrapAttributesThreshold": -1,
  "pugFramework": "auto",
  "pugEmptyAttributes": "none",
  "pugEmptyAttributesForceQuotes": [],
  "pugClassNotation": "literal",
  "pugIdNotation": "literal",
  "pugClassLocation": "before-attributes",
  "pugExplicitDiv": false,
  "pugPreserveAttributeBrackets": false,
  "pugPreserveWhitespace": false,
  "pugClosingBracketIndentDepth": 0,
  "pugBracketSpacing": true,
  "pugBracketSameLine": false,
  "pugPrintWidth": 100,
  "pugSemi": true,
  "pugSingleQuote": false,
  "pugTabWidth": 2,
  "pugUseTabs": false,
  "pugArrowParens": "always"
}

```

## Input

```pug
.test
  +btn({ color: "red" })

  +btn({
    color: "red",
    color1: "red",
    color2: [
      {
        color3: "red",
      },
      {
        color4: "red",
      },
    ],
    color5: "red",
    color6: "red",
    color7: "red",
    color8: "red",
    color9: "red",
    color10: "red",
    color11: "red",
    color12: "red",
  })

  .test
    .test
    +tools-group({
      icon: "/static/tool.svg",
      label: "tool",
      title: "Tool title",
      tools: [
        {
          icon: "/static/tool-item-1.svg",
          title: "Tool item title",
        },
        {
          icon: "/static/tool-item-2.svg",
          title: "Tool item title",
        },
        {
          icon: "/static/tool-item-3.svg",
          title: "Tool item title",
        },
        {
          icon: "/static/tool-item-4.svg",
          title: "Tool item title",
        },
      ],
    })
```

## Output or Error

```bash
.test
  +btn({ color: "red" })

  +btn({ color: "red", color1: "red", color2: [{ color3: "red" }, { color4: "red" }], color5: "red", color6: "red", color7: "red", color8: "red", color9: "red", color10: "red", color11: "red", color12: "red" })

  .test
    .test
    +tools-group({
  icon: "/static/tool.svg",
  label: "tool",
  title: "Tool title",
  tools: [
    { icon: "/static/tool-item-1.svg", title: "Tool item title" },
    { icon: "/static/tool-item-2.svg", title: "Tool item title" },
    { icon: "/static/tool-item-3.svg", title: "Tool item title" },
    { icon: "/static/tool-item-4.svg", title: "Tool item title" },
  ],
})
```

## Expected Output

```pug
.test
  +btn({ color: "red" })

  +btn({
    color: "red",
    color1: "red",
    color2: [
      {
        color3: "red",
      },
      {
        color4: "red",
      },
    ],
    color5: "red",
    color6: "red",
    color7: "red",
    color8: "red",
    color9: "red",
    color10: "red",
    color11: "red",
    color12: "red",
  })

  .test
    .test
    +tools-group({
      icon: "/static/tool.svg",
      label: "tool",
      title: "Tool title",
      tools: [
        {
          icon: "/static/tool-item-1.svg",
          title: "Tool item title",
        },
        {
          icon: "/static/tool-item-2.svg",
          title: "Tool item title",
        },
        {
          icon: "/static/tool-item-3.svg",
          title: "Tool item title",
        },
        {
          icon: "/static/tool-item-4.svg",
          title: "Tool item title",
        },
      ],
    })
```

## Reproduction

[plugin-pug-formatting.zip](https://github.com/user-attachments/files/22990487/plugin-pug-formatting.zip)

1. unzip
2. npm install
3. open text.pug
4. format document

LucasLeao18:
## Summary

- format mixin-call arguments against the remaining Pug line width instead of the internal 9000-column interpolation width
- re-indent multiline JavaScript output relative to the Pug call
- cover long and nested mixin calls with regression fixtures

Fixes #620.

## Verification

- `pnpm run format`
- `pnpm run lint`
- `pnpm run build`
- `pnpm vitest run` (142 files, 326 tests)
- `git diff --check`

`pnpm run ts-check` was also run; it currently fails in unchanged `eslint.config.ts` and `vitest.config.ts` because the locked dependency graph contains incompatible `@eslint/core` type versions and a coverage-option type mismatch. The build, linter, and full test suite pass.


Shinigami92:
CI is green 👍 Need to review it when I find time

src/printer.ts
```diff
@@ -2194,9 +2197,22 @@ export class PugPrinter {
       args = await format(`x(${args})`, {
         parser: 'babel',
         ...this.codeInterpolationOptions,
+        printWidth: Math.max(this.options.pugPrintWidth - callPrefixWidth, 1),
```
Shinigami92: **issue (blocking):** mixin call inside `#[...]` now emits invalid Pug

Pug interpolation must stay on a single line. With `printWidth: 9000` that was guaranteed; now the arguments can wrap inside the interpolation and the result no longer parses.

Input (`pugPrintWidth: 60`):

```pug
p
  | text #[+btn({ color: "red", color1: "red", color2: "red", color3: "red", color4: "red" })] more
```

Output on this branch:

```pug
p
  | text #[+btn({
    color: "red",
    color1: "red",
    color2: "red",
    color3: "red",
    color4: "red",
  })] more
```

Re-formatting that output throws:

```
Pug:2:18
  > 2|   | text #[+btn({
------------------------^
The end of the string reached with no closing bracket ) found.
```

On `main` the same input is stable and idempotent (it stays on one line). So this is a regression introduced by the PR, and it is a silent corruption: the user's template is left unparseable.

`currentlyInPugInterpolation` (`src/printer.ts:229`) is already set by the time `call()` runs and is used for exactly this kind of guard elsewhere (`src/printer.ts:877`, `src/printer.ts:990`). Suggested fix - keep the old unbounded width, and skip the re-indent, while inside an interpolation:

```ts
printWidth: this.currentlyInPugInterpolation
  ? this.codeInterpolationOptions.printWidth
  : Math.max(this.options.pugPrintWidth - callPrefixWidth, 1),
```

src/printer.ts
Shinigami92: **issue (non-blocking):** `currentLineLength` is inflated by the multiline result

```ts
this.currentLineLength += result.length;
```

`result` may now span many lines, but `currentLineLength` is meant to track the _current_ line. It is only reset on `newline`/`indent`/`outdent`, so everything printed after a wrapped call on the same logical line sees an inflated value. The practical effect is the `currentLineLength > pugPrintWidth` check at `src/printer.ts:990`, which then force-wraps attributes that would have fit:

Input (`pugPrintWidth: 60`):

```pug
.test
  +btn({ color: "red", color1: "red", color2: "red", color3: "red", color4: "red" })(data-a="1" data-b="2")
```

Output:

```pug
.test
  +btn({
    ...
  })(
    data-a="1",
    data-b="2"
  )
```

`  })(data-a="1" data-b="2")` is 27 characters — comfortably under 60 — so the attribute wrapping is unnecessary. The output is still valid and idempotent, so this is cosmetic, not a correctness bug. Adding only the last line's length would be more accurate:

```ts
this.currentLineLength += result.slice(result.lastIndexOf('\n') + 1).length;
```

Fine as a follow-up if you'd rather not touch shared state here — but worth noting somewhere, because it only becomes reachable _because_ of this change.

Shinigami92: **issue (non-blocking):** the width budget is off by one

Prettier formats the wrapper `x(${args})`, so its own measurement already includes one character (`x`) standing in for the call prefix. Subtracting the full `callPrefixWidth` on top of that double-counts that column, so calls break one character earlier than `pugPrintWidth` requires.

Reproduction — `+b({ a: "x" })` is exactly 14 characters:

| `pugPrintWidth` | result             |
| --------------- | ------------------ |
| 14              | wraps (should fit) |
| 15              | stays on one line  |

`Math.max(this.options.pugPrintWidth - callPrefixWidth + 1, 1)` fixes it. Worth a dedicated boundary test, since this is the kind of thing that silently drifts.

src/printer.ts
```diff
@@ -2186,6 +2186,9 @@ export class PugPrinter {
 
   private async call(token: CallToken): Promise<string> {
     let result: string = `${this.computedIndent}+${token.val}`;
+    const callIndent: string = this.indentString.repeat(this.indentLevel);
+    const callPrefixWidth: number =
+      this.indentLevel * this.options.pugTabWidth + token.val.length + 1;
```
Shinigami92, line 2191: **nitpick (non-blocking):** continuation lines over-reserve the mixin-name width

Continuation lines only receive `callIndent`, yet the width budget was reduced by `token.val.length + 1` for _every_ line. For a long mixin name the continuation lines wrap noticeably earlier than needed. Reserving the full prefix only for the first line is the exact model; the current approximation is conservative and acceptable, but it deserves a one-line comment explaining the trade-off rather than leaving the reader to reverse-engineer it.

tests/mixins/multiline-calls.pug
Shinigami92, line 1: **suggestion (non-blocking):** make the new fixture pair actually unformatted → formatted

`multiline-calls.pug` and `multiline-calls.formatted.pug` have identical content, so the new test only asserts **idempotency**. It would pass on any implementation that happens to leave already-correct input alone; it does not demonstrate that the reported breakage is repaired.

The repo convention elsewhere is `unformatted.pug` → `formatted.pug` (see `tests/mixins/unformatted.pug`, and the `source:`/`target:` pairs in `tests/options/**`). Please make the source the actual broken output from #620:

```pug
.test
  +btn({ color: "red", color1: "red", color2: [{ color3: "red" }, { color4: "red" }], color5: "red", color6: "red", color7: "red", color8: "red", color9: "red", color10: "red", color11: "red", color12: "red" })

  .nested
    +tools-group({
  icon: "/static/tool.svg",
  label: "tool",
})
```

Keeping the current identical pair as a second "already formatted stays formatted" case is still worthwhile - just not as the only assertion.

tests/mixins/mixins.test.ts
```diff
@@ -6,4 +6,13 @@ describe('Mixins', () => {
     const { expected, actual } = await compareFiles(import.meta.url);
     expect(actual).toBe(expected);
   });
+
+  it('should indent multiline call arguments relative to the call', async () => {
+    const { expected, actual } = await compareFiles(import.meta.url, {
+      source: 'multiline-calls.pug',
+      target: 'multiline-calls.formatted.pug',
+      formatOptions: { pugPrintWidth: 100 },
```
Shinigami92, line 17: **note (non-blocking):** the test pins a non-default `pugPrintWidth`

`pugPrintWidth: 100` while the default is 80. That is fine and matches the config in #620, but a short comment saying _why_ would help the next reader, since the fixture layout depends on it.

src/printer.ts
```diff
@@ -2194,9 +2197,22 @@ export class PugPrinter {
...
+        args = lines
+          .map((line, index) => {
+            if (index === 0) {
+              return line;
+            }
+
+            return callIndent + line;
+          })
+          .join('\n');
+      }
```
Shinigami92: **nitpick (non-blocking):** redundant guard in the re-indent block

```ts
const lines: string[] = args.split('\n');
if (lines.length > 1) {
  args = lines.map((line, index) => (index === 0 ? line : callIndent + line)).join('\n');
}
```

The `lines.length > 1` guard is redundant - a single-line string maps to itself. The equivalent logic in `start-pipeless-text` (`src/printer.ts:2080-2086`) is written as a plain `split`/`map`/`join` without a guard; matching that keeps the file consistent. Consider also preserving empty lines as empty (that handler does `line ? indent + line : ''`), though Prettier's output for a call expression is unlikely to contain blank lines in practice.

tests/mixins/mixins.test.ts
Shinigami92, line 1: **todo (blocking):** cover the interpolated mixin call in the fixtures

It is the one shape the current tests do not exercise at all, and the shape that regresses. Worth adding alongside whatever guard lands for the issue above.

LucasLeao18:
Addressed the review in fba757b:

- preserve the unbounded code-interpolation width and skip re-indentation for mixin calls inside `#[...]`
- track only the final rendered line in `currentLineLength`
- correct the one-column width budget and add 13/14/15-column boundary coverage
- turn the multiline fixture into an actual unformatted-to-formatted case
- add interpolation and trailing-attribute regression fixtures, including idempotency checks
- document the conservative continuation-line width trade-off and the `pugPrintWidth: 100` fixture choice

Verified locally: all 142 test files / 330 tests pass, and ESLint passes.

src/printer.ts
LucasLeao18: Addressed in fba757b: the code now includes a one-line comment documenting that the prefix width is conservatively reserved on every line.

tests/mixins/multiline-calls.pug
LucasLeao18: Addressed in fba757b: multiline-calls.pug now contains the broken/unformatted input, while multiline-calls.formatted.pug remains the expected output. The separate idempotency case is also retained.

Shinigami92 (approved):

Shinigami92:
I need to setup potentially some more stuff before doing a new release, and I need to check if the infra rework had made effective changes to the dist. So you might need to wait until weekend before I do a release. I just want to do it safe and not rushed.

